PR #11, opened 2026-08-06
Verify Nextcloud 30–34 support with a cross-version × DB CI matrix

aktasfatih:
## What this does

`info.xml` / README already declare **Nextcloud 30–34**, but CI, the dev stack
and release all booted **NC 34 + Postgres only** — so the 30–34 claim was
declared but never *validated*. This PR adds the cross-version validation and a
static compatibility audit, without over-claiming: the actual per-combo verdict
comes from this PR's CI run.

## The matrix it adds

A new `install-matrix` CI job boots a throwaway Nextcloud per combo, enables
Kanso (running every migration), runs `occ upgrade`, asserts the schema was
created, and runs PHPUnit against that version's bundled PHP:

| NC \ DB | SQLite | Postgres | MariaDB |
|--------|:------:|:--------:|:-------:|
| 30 | ✅ | ✅ | — |
| 31 | ✅ | ✅ | — |
| 32 | ✅ | ✅ | — |
| 33 | ✅ | ✅ | — |
| 34 | ✅ | ✅ | ✅ |

**11 install combos.** The full Playwright **e2e suite stays on NC 34 + Postgres
only** (running ~170 serial browser tests across all combos would be
prohibitively slow). SQLite runs on every version because it needs no service
container and it's the backend that surfaced the original migration bug.

## How each combo is validated

Per (version, db): build frontend → boot stack (`dev/setup.sh`, now
parametrized by `NC_VERSION` / `KANSO_DB`) → `occ upgrade` → **`dev/smoke.sh`**
(new) → PHPUnit in-container.

`dev/smoke.sh` is the load-bearing check: it asserts Kanso is enabled, every
migration file is recorded applied in `oc_migrations`, and the `kanso_*` tables
physically exist (across sqlite/postgres/mysql) — i.e. it catches the class of
bug where an over-long default-named index silently creates **zero** tables.
Validated locally end-to-end against the running NC 34 + Postgres stack
(42 migration files, 45 recorded, 31 tables, all core tables present).

## Dev stack parametrization

`dev/docker-compose.yml` + `dev/setup.sh` now take:
- `NC_VERSION` → `nextcloud:<version>-apache` (default 34)
- `KANSO_DB` → `postgres` (default) | `mysql` (mariadb:11) | `sqlite` (no db service)

selected via compose profiles. Local `./setup.sh` behaviour is unchanged
(NC 34 + Postgres + notify_push). `dev/.db.env` is generated per-run and
gitignored.

## Compatibility audit (static)

Two focused audits were run over `lib/`; **no code fixes were required**:

- **Migrations (all 42 files):** every `addIndex`/`addUniqueIndex` is explicitly
  named; longest name is 27 chars (under the 30-char DBAL hard limit); no
  default-named indexes, no foreign keys, no raw/dialect-specific SQL. Data
  backfills use the portable QueryBuilder only. Boolean columns follow the
  NULL-safe NC pattern. The original "default-named PK over the limit" pitfall
  is fully addressed. (7 index names sit in the 24–27 char range — above the
  stricter 23-char self-imposed budget but well under the hard limit; left as-is
  to avoid churning index names, which would itself require migrations.)
- **OCP/OC APIs:** clean for NC 30–34. Modern `TimedJob`, PSR `LoggerInterface`,
  `IEventListener`, routing attributes (`@since 27`), `UnknownNotificationException`
  (`@since 30` — safe at min-version 30). No `\OC::$server`, no legacy Symfony
  dispatcher. `#[\Override]` is always fully-qualified (inert on PHP 8.2).
  sabre/vobject 5.0 API already handled (nullable `current()`).
- **PHP / info.xml consistency:** app targets PHP 8.2+; `composer.json` pins
  `platform.php = 8.2.27`; no PHP 8.3-only stdlib used. Consistent.

## info.xml / README

Already `min-version="30" max-version="34"` on `main` and README already reads
"30–34" — so no change was needed there. This PR adds a CHANGELOG entry for the
matrix and the dev-stack knobs, plus README examples for `NC_VERSION`/`KANSO_DB`.

## Honest status / what needs a human eye

- **Verified locally:** NC 34 + Postgres install + smoke (real run). YAML,
  shell, and compose profile resolution lint clean.
- **Pending CI:** the other 10 combos (NC 30–33 × sqlite/postgres, NC 34 ×
  mariadb). NC 30–33 ship PHP 8.2/8.3; the in-container PHPUnit uses
  `composer.json`'s platform override so resolution is version-independent, but
  the real runtime verdict is this PR's CI.
- **Scoped follow-up (not a blocker):** MariaDB is covered on NC 34 only, to
  avoid a 5× combo blow-up. Widen it across versions if a MySQL-dialect issue
  ever surfaces.

---

### CI findings (updated after live matrix runs)

The matrix did its job and caught two real cross-version issues, now fixed in this PR:

1. **Migration bug (NC 30–32): over-long default primary-key names.** On NC 30–32,
   several tables' default-generated PRIMARY KEY name exceeded Nextcloud's
   23-char index-name limit, aborting app install with *"Primary index name on
   'oc_kanso_comment_reactions' is too long"* — zero tables created. Fixed by
   giving 13 at-risk tables explicit short PK names. **Confirmed fixed:** NC 32
   (sqlite + postgres) and NC 33/34 now pass the install + migration + schema
   smoke check.

2. **Older-image install quirks (NC 30/31).** Two environment issues on the
   self-hosted runner, both fixed: (a) older NC images' first-boot install can
   exceed the original 7.5-min wait — raised to 15 min; (b) NC 30/31's installer
   aborted with *"Cannot write into 'apps' directory"* — added a pre-installation
   hook that makes the bundled apps dir writable.

**Latest confirmed-green combos:** NC 32 (sqlite+postgres), NC 33 (postgres),
NC 34 (postgres). NC 30 (both DBs) and NC 31 are being re-validated by the CI
run that includes the apps-writable + timeout fixes. e2e (NC 34 + postgres,
~1.5h) runs separately. See the PR checks for the authoritative per-combo result.

The in-container PHPUnit step was removed from the matrix: the ARC runner's dind
network can't reach the internet from inside a container, so it can't fetch
composer to install the gitignored dev deps. PHPUnit remains covered by the
dedicated `unit-php` job across PHP 8.2 and 8.3.


aktasfatih:
## CI matrix results — NC 32/33/34 green; NC 30 needs a decision

After several fix rounds (each addressing a distinct, log-confirmed root cause), the matrix has settled:

| NC \ DB | SQLite | Postgres | MariaDB |
|--------|:------:|:--------:|:-------:|
| 30 | ❌ | ❌ | — |
| 31 | ⚠️ flaky | ⚠️ flaky | — |
| 32 | ✅ | ✅ | — |
| 33 | ✅ | ✅ | — |
| 34 | ✅ | ✅ | ✅ |

**What the matrix caught and this PR fixed (all confirmed green on 32/33/34):**
1. **Migration bug — over-long default PK names on NC 30–32** (the headline find). Enabling Kanso aborted install with *"Primary index name on 'oc_kanso_comment_reactions' is too long"* → zero tables. Fixed by naming 13 tables' PKs. Confirmed: NC 32 sqlite+postgres now install cleanly (this was the first failure).
2. **Postgres install race** — making the DB configurable dropped the NC→DB start dependency; NC raced Postgres startup. Fixed with DB healthchecks + `depends_on: service_healthy`. Confirmed: NC 32/33/34 postgres now stable.
3. **Boot timeout** raised 7.5→15 min for slow cold-image installs.

**Needs a human decision — NC 30 (and intermittently NC 31):**
NC 30's installer consistently aborts with **"Cannot write into 'apps' directory"** on this runner, on **both sqlite and postgres** (so it's not the DB race). A pre-installation hook that `chown`/`chmod`s `/var/www/html/apps` **runs successfully but does not satisfy NC 30's check** — this is an NC-30-specific installer behavior with no clean shim found. NC 31 shows the same message intermittently; NC 32+ never do.

Options (pick one):
- **(a) Pre-seed `appstoreenabled=false`** via an install-time config so the writability check is skipped (needs verifying it doesn't perturb the app's own install).
- **(b) Mount a dedicated writable `apps` volume / adjust the base image perms** so NC 30's exact `is_writable()` check passes.
- **(c) Narrow the floor to NC 31 or 32** (`info.xml min-version`, README, matrix) if NC 30 support isn't worth the CI contortions — the code itself is NC-30-clean (the static audit found no NC-30-absent APIs); this is purely a first-boot install-check quirk of the NC 30 *image* on this runner.
- **(d) Treat NC 30 as "declared, best-effort"** — keep min-version=30 but mark the NC 30 matrix leg `continue-on-error`.

I did not guess between these because it's a product/scope call. Everything else (info.xml 30–34, README, CHANGELOG, the matrix, the migration + race fixes) is done and green on 32–34.

PR #75, opened 2026-08-23
test: share e2e helpers and make the suite parallel-ready

aktasfatih:
## What

Grooming pass on the e2e suite (Deck card #4143). Two problems addressed:

1. **Redundancy** — every one of the ~128 spec files copy-pasted its own
   `BASE`/`AUTH`/`HEADERS` constants plus `ncLogin()` and an
   `apiGet/apiPost/apiDelete` cluster (116× `ncLogin`, 40× `apiPost`, …).
2. **No isolation → forced serial** — the suite runs `workers: 1` because every
   spec acts as the same `admin` user against one shared DB (fixed board names,
   delete-then-recreate, and per-user aggregate views like My Work / Inbox /
   Search all collide if two specs run concurrently). That serial constraint is
   what makes the full `e2e` check ~1.5h.

## How

- **New `tests/e2e/helpers.js`** — one canonical `api` client
  (`get/post/patch/put/delete/send/raw`), `ncLogin`, `gotoBoard`/`boardUrl`,
  `authFor`, `makeApi`, `provisionUser`/`deleteUser`.
- **126 specs repointed** at it, **behaviour-preserving** (same admin user, same
  API calls, same steps). Error-swallowing cleanups (`.catch(() => {})`),
  bespoke non-throwing `{ok,status,body}` clients, DAV/upload helpers, and
  multi-user `test.use` storageState opt-outs were all preserved as-is.
- **Parallel-ready seam** — the exported `test` is extended with a worker-scoped
  `user`/`api` fixture. With `E2E_ISOLATE=1` each Playwright worker provisions
  its own Nextcloud user (`kansoe2e_w<n>`), namespacing board lists + aggregate
  views per worker so `E2E_WORKERS` can be raised. Flag **off by default** → the
  fixtures resolve to `admin` and behaviour is byte-for-byte the current suite.
- `playwright.config.js`: `workers` now env-driven (`E2E_WORKERS`), default 1.
- `package.json`: `test:e2e:parallel` (`E2E_ISOLATE=1 E2E_WORKERS=4`).
- `tests/e2e/README.md`: documents the helpers + the isolation/parallel path.
- `deck-import` and `share-from-files` keep their bespoke non-mappable clients
  (multi-base-URL / non-throwing) and are intentionally left as-is.

## Scope note

This PR is the **foundation** — dedup + parallel-*ready*, still `workers: 1`, so
the required `e2e` check stays green. Actually flipping the default to parallel
[...]

PR #76, opened 2026-08-23
test: run the e2e suite in parallel with per-worker Nextcloud users

aktasfatih:
Follow-up to the e2e helpers PR (#75). That one made the suite parallel-*ready*
but nothing actually consumed the isolation. This one makes it **real** and
turns it on in CI.

## Under `E2E_ISOLATE=1`, each Playwright worker is its own Nextcloud user
- **Browser session:** a per-worker `storageState` — pages start logged in as
  `kansoe2e_w<n>`, not the shared admin.
- **API identity:** `helpers.js` rebinds the live `api` / `me` / `currentAuth`
  bindings per worker, so a spec's `import { api }`, its self-refs
  (`/assignees/${me}`), and its bespoke `fetch` clients all act as the worker
  user — **no per-spec plumbing** for the common case (ES live bindings + an
  `auto` worker fixture that runs before any `beforeAll`).
- **Second identity:** a new `peer` worker fixture (off-isolate = the dev
  `tester`; on-isolate = a per-worker peer) for board-sharing / ACL / peer-login
  specs. Specs that provisioned their own extra users now suffix the worker
  index so parallel workers never collide.

With `E2E_ISOLATE` **unset** (the default) every binding resolves to admin /
tester and behaviour is byte-for-byte the serial suite — dormant, not a code
path specs must reason about.

## CI
The `e2e` job now runs `E2E_ISOLATE=1 E2E_WORKERS=2` — conservative on purpose:
the self-hosted runner hosts the whole stack in-pod and is ~4-5x slower than a
dev box, so over-subscribing php-fpm/postgres would trade wall-clock for timeout
flakiness. `E2E_WORKERS` can be raised once the runner has headroom.

## Validation (local, `E2E_ISOLATE=1 --workers=4`)
Ran every spec category in parallel:

| category | result |
|---|---|
| self-user (`me`) | 40/40 |
| bespoke clients (`currentAuth`) | 42/45 |
| multi-user API (`peer`) | 38/40 → 40 after fix |
| browser-peer login | 17/17 |
| provisioning / anon | 19/19 |

Every non-pass reproduces **identically as admin** on the dev stack (pre-existing
purge-endpoint 500 / contacts indexing / the known `trash` flake) — none are
isolation regressions. Two-user probe confirmed each worker's `beforeAll` API
client *and* browser `whoami` resolve to its own user.

Diff is helpers.js + the isolation fixture, a docs update, the CI env, and
mechanical per-spec swaps (`'admin'`→`me`, tester→`peer`, bespoke-client auth→
`currentAuth`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,15 +255,17 @@ jobs:
 
       # Run the suite in PARALLEL: each Playwright worker provisions its own
       # Nextcloud user (E2E_ISOLATE) so fixed board names and per-user aggregate
-      # views are namespaced per worker. Conservative worker count — the runner
-      # hosts the whole stack in-pod and is ~4-5x slower than a dev box, so
-      # over-subscribing php-fpm/postgres would trade wall-clock for timeout
-      # flakiness. Raise E2E_WORKERS once the runner has headroom.
+      # views are namespaced per worker. The runner hosts the whole stack in-pod;
+      # its `runner` container is capped at 4 CPU (node talos-worker-3 has 48
+      # cores, ~8% used). At workers=2 the run took ~30m using ~1.3 of 4 cores,
+      # so 4 workers fits the cap. If timeout flakiness appears, the bottleneck
+      # is the single php-fpm/postgres backend (tune pm.max_children /
+      # max_connections) or the 4-CPU cap — not the node.
       - name: Run e2e tests
         run: npm run test:e2e
         env:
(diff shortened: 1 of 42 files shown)


aktasfatih:
:tada: This pull request is included in version 0.15.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.15.0).

PR #79, opened 2026-08-24
fix: Deck import with duplicate assignments + Timeline perf on wide date ranges

aktasfatih:
Two independent, user-facing fixes.

## fix: Deck import — boards that assign a label or member to a card twice
A Deck board where the same label or assignee is stored against a card more than once (Deck allows this — `deck_assigned_labels` / `deck_assigned_users` have no uniqueness constraint) failed to import **entirely**: Kanso's `kanso_card_labels_uniq` / `kanso_card_assign_uniq` reject the duplicate, and because the whole import runs in one transaction the constraint violation rolled everything back — no board was created.

- Fix: `DeckReader` now deduplicates label ids and assignee uids **per card** as they're read (`dedupePerCard`, order-preserving), so a duplicated Deck row can never reach Kanso's unique constraint.
- Reproduced against a real Deck board carrying several twice-assigned labels; such boards now import cleanly.
- Regression tests added in `tests/unit/Service/DeckReaderTest.php`.

## fix: Timeline (Gantt) stays fast when a card spans a huge date range
The Timeline rendered an absolutely-positioned track sized to the raw min-start / max-due across **all** cards, so a single outlier (e.g. 2018→2030) inflated `trackWidth` and the per-day/tick/weekend DOM loops → jank on layout, scroll and re-render.

- Fix: render a bounded **window** (today −6mo … +12mo) instead of the raw domain; feed the window into `totalDays`/ticks/month-ticks/weekend-bands/`trackWidth`. Bar geometry stays anchored to real dates and is clipped by the existing `overflow-x:auto`.
- Outliers are **indicated, not dropped**: an "extends earlier / later" edge marker appears when data falls outside the window.
- Small/normal boards whose data already fits the window render byte-identical (fallback to raw bounds). No trackWidth hard-cap, no virtualization, no new settings.
- Perf e2e added to `tests/e2e/timeline-view.spec.js`: a many-year card must keep the scroll container bounded and show the edge affordance.

### DIFF
--- a/lib/Service/DeckReader.php
+++ b/lib/Service/DeckReader.php
@@ -150,7 +150,12 @@ class DeckReader {
 	}
 
 	/**
-	 * label-id lists keyed by card id, for the given cards.
+	 * label-id lists keyed by card id, for the given cards. Deduped per card:
+	 * Deck's `deck_assigned_labels` has no unique constraint on
+	 * (card_id, label_id), so the same label can be stored against a card twice.
+	 * Kanso's `kanso_card_labels_uniq` does NOT allow that, so a duplicate row
+	 * would abort the whole (single-transaction) import - dedupe it here at the
+	 * source, order-preserving.
 	 *
 	 * @param int[] $cardIds
 	 * @return array<int, int[]>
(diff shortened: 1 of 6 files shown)


aktasfatih:
:tada: This pull request is included in version 0.16.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.16.0).

PR #81, opened 2026-08-24
feat: recurring cards use a start–end window that slides forward (fixes date resets)

### LINKED ISSUE #80: [Bug] Start date of new recurring tasks automatically resets to creation date (related to #65)
Hi @aktasfatih,

I encountered another bug regarding the dates of recurring tasks. It seems somewhat related to the previous issue we had (#65), but this time without any duplication.

Description of the bug:
Yesterday, I created 3 new tasks. I manually set a specific Start Date for them and set the repeat interval to "Yearly". Today, I noticed that the system automatically overwrote the Start Date of all three tasks to yesterday's date (August 23, 2026). The tasks were not duplicated (so the fix for #65 seems to work!), but the custom date was still modified in the background.

**To Reproduce:**
Create a new task.
Manually set a Start Date (different from the current day).
Set the repeat interval to "Yearly" (or any other interval).
Check back later (likely after a background automation/cron job runs).

Notice that the custom Start Date has been forcefully overwritten and reset to the date the task was originally created.

**Expected behavior:**
The manually defined Start Date should be saved permanently and must never be overwritten by the system in the background.

Could you check if the date calculation logic in the background job is accidentally overwriting the initial start date? Thanks for your help!

aktasfatih:
## Summary

Recurring cards had a cluster of date bugs that kept resurfacing. This branch fixes the root cause and reworks how a repeating card's dates behave so the model is predictable and self-explanatory.

### Bug fix
Turning on Repeat scheduled the rule's first run for the moment it was created, so the next background pass (~15 min later) fired it immediately and reset the card's date to its creation day — a "Yearly" repeat was re-stamped to today within minutes. The same off-by-one meant that editing a repeat to a more frequent cadence didn't take effect until the old, far-off date arrived. Both paths now arm the next run strictly in the future, so nothing fires on the very next pass and a cadence change takes effect from its next occurrence, while preserving the existing anti-duplication guarantee.

### Recurring dates now behave as a start → end window
A card is a **Start → End** window, and a repeat slides that whole window forward to each occurrence, keeping its length, instead of stamping a single date and ignoring the start date. The schedule anchors on the card's **Start date** (so "starts on the 5th, repeats weekly" lands on the 5th, 12th, 19th; a Start set for the future fires on that date), and editing the Start/End date re-points the series so it follows the new dates.

### Terminology and guidance
The card's second date is now labelled **"End date"** (was "Due date") everywhere it faces the user, so a card reads as a Start → End window rather than a task with a hard deadline. Small ⓘ tips next to the Start date, End date and Repeat fields explain each one in plain language. The database column is unchanged, so this is a label-only change with no migration.

## Tests
- **Unit:** recurrence and card-service coverage for the anchor, the window slide, the re-anchor-on-edit, and the cadence-change path.
- **e2e:** new deterministic regression guards for each bug, checked via the rule's cached next-fire time and create-now, so nothing waits on the background job.

Closes #80.

### DIFF
--- a/lib/Db/RecurRuleMapper.php
+++ b/lib/Db/RecurRuleMapper.php
@@ -108,6 +108,24 @@ class RecurRuleMapper extends QBMapper {
 	}
 
 	/**
+	 * Every recurrence rule anchored on a template card. Used to re-point a card's
+	 * repeat schedule when its Start/End date is edited, so the series follows the
+	 * new dates (what a user naturally expects when rescheduling a repeating card).
+	 * A card carries at most a handful of rules, so no ordering is needed.
+	 *
+	 * @return RecurRule[]
+	 * @throws Exception
+	 */
+	public function findByTemplateCard(int $cardId): array {
+		$qb = $this->db->getQueryBuilder();
+		$qb->select('*')
+			->from($this->getTableName())
+			->where($qb->expr()->eq('template_card_id', $qb->createNamedParameter($cardId, IQueryBuilder::PARAM_INT)));
+
+		return $this->findEntities($qb);
+	}
+
+	/**
 	 * Removes every recurrence rule anchored on a template card - the cascade for
 	 * a card purge. A rule whose template is hard-deleted can never spawn again
 	 * (its template read throws), so purging the card must drop its rules too;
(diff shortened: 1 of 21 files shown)


aktasfatih:
:tada: This pull request is included in version 0.16.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.16.0).

PR #88, opened 2026-08-26
docs(claude): dogfood Kanso for this repo's own board block

aktasfatih:
## Why

Kanso's own development board moved off Deck onto Kanso — dogfooding. Since the Deck→Kanso import reassigned every board / stack / label id, the block in `CLAUDE.md` was pointing `/pm`, `/work` and `/create-tasks` at ids that no longer exist.

The failure is silent, not loud: the `nextcloud` MCP's `deck_*` tools still return `200`, just with the old Deck boards.

## What changed

- Board access documented as the **`kanso` MCP** (`mcp__kanso__kanso_*`) — this repo's own `mcp/` server.
- Board id pinned and verified against the live API (**19**, was 14).
- Column and label ids resolved by name at runtime rather than hardcoded. One `kanso_get_board` call returns stacks, labels and all card summaries, so name-matching costs nothing extra, and those are the ids most likely to drift as the board is reshaped.
- Ordering language updated: `sortKey`, not a numeric `order`.

Docs-only: no app code, no schema, no CI. Typed `docs` so semantic-release does not cut a version.

> Committed with `--no-verify`: a fresh `git worktree` has no `node_modules`, so the husky `commit-msg` hook could not launch commitlint. The message was validated separately against this repo's own commitlint config (`./node_modules/.bin/commitlint` from the main checkout) and passes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -23,9 +23,20 @@ performance-first kanban board. AGPL, targets Nextcloud 30–34.
 - NEVER test against prod (sv1). Use the local docker dev stack in `dev/` (copied from ../deck_recurrence/dev). Prod only gets tagged releases via the scoped deploy playbook.
 - The PM charter lives in `.claude/pm-charter.md` — goal, stage, non-goals, ship bar. Respect its over-engineering guard: this is an MVP; no speculative abstractions.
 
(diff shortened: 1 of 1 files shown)


aktasfatih:
Superseded: landing this directly on main instead of via PR.

PR #89, opened 2026-08-26
feat: App Store polish, per-board card sections, timeline dependencies and two data-safety fixes

aktasfatih:
Sprint `sprint-2026-08-26` — **ship the App-Store-facing polish and close two verified defects**. Eight cards, each validated against the code before implementation, implemented and tested in isolation, then committed onto this one branch.

## What's in here

| --- | --- | --- |
| `ee53980` | #9843 | German + French `<summary>`/`<description>` for the App Store listing. `<name>` stays "Kanso"; German terminology cross-checked against the existing `de` catalogue. |
| `b5edae1` | #5891 | **Bug**: a card could never be dropped into an empty column in list view. The group-drop overlay carried `pointer-events: none`, so native drag hit-testing could never resolve it — structurally impossible, not flaky. Now pointer-active only while a drag is in flight. |
| `d49a644` | #9847 | **`appinfo/info.xml` did not validate against the App Store schema** (`<background-jobs>` out of sequence) — pre-existing on `main`, and a live risk of the store rejecting the upload. Plus a CI guard so it can't drift again. |
| `9d34fca` | #9845 | **Data loss**: two people editing one description meant the second save silently destroyed the first. Now the editor sends the version it was seeded from and a stale write is refused with `409 description_conflict`, showing both versions with *Keep mine* / *Use theirs*. |
| `3dbc011` | #9844 | CI now lints every `.po` for syntax and placeholder consistency, so a renamed `{count}` or dropped `%1$s` can't ship broken in a locale nobody here runs. |
| `52a66e6` | #5885 | Drop a card on another card's centre band to make it a sub-card, in both kanban and list views, with an affordance distinct from reordering. |
| `125e1e6` | #5894 | Board managers can hide card sections their team never uses (contacts, attachments, GitHub, time tracking, cover colour). One nullable JSON column, `NULL` = all enabled, no backfill. |
| `1594740` | #5896 | Blocking dependencies render as elbow arrows on the timeline, **red when violated** (blocked card starts before its blocker finishes). Board-scoped edge list with whole-edge visibility masking. |

## Please look closely at these

Three deliberate decisions that a reviewer should agree with rather than discover:

1. **`125e1e6` widens two public-share payload leak-guards** (PHPUnit + e2e) so the anonymous board payload can carry the new `cardFeatures` map. The map is person-free booleans, but these guards exist to stop data reaching anonymous viewers — worth a second pair of eyes.
2. **`125e1e6` gates the five card features client-side only.** The API still accepts an attachment upload or timer start on a board with the switch off. Deliberate and documented in `CardFeatures`' docblock: re-enabling must return intact data, and webhook intake / imports / MCP / CalDAV must not break because a manager tidied a card modal.
3. **`52a66e6` ships nesting but NOT drag-to-un-nest**, against the card's original spec. On a kanban board a sub-card looks like any other tile, so "drag this subtask to the next column" would have silently cleared its parent with no undo. Un-nesting stays an explicit card-panel action; the safe drag version is carded for list view only.

## Behaviour changes worth knowing

- **`b5edae1`**: with the overlay live during a drag, dropping in the gap *below* a group's last card now lands on the next group's header and appends there. Dropping on a card's bottom edge still appends within the group.
- **`9d34fca`**: an existing e2e that asserted last-writer-wins was updated — it encoded the lossy behaviour as correct.

## Verification

Every card ran its own tests before committing; the full suite is CI's job. Highlights: PHPUnit 1,442 green (whole suite, in-container) after `125e1e6`; psalm and php-cs-fixer clean throughout; new e2e specs for empty-column drop, description conflict, nesting, card features and timeline dependencies, each run green — the list-view drop fix was additionally negative-tested by reverting the fix and confirming the new spec fails.

`<version>` is untouched in every commit — version bumps are release-only. The `125e1e6` migration lands unbumped by design.

## Not shipped, carded instead

#9848 (revision-token hardening for the description conflict guard), #9849 (editing presence — needs new server infrastructure), #9850 (safe list-view un-nesting), #9851 + #9852 (local e2e environment gaps that make developers ignore red).


aktasfatih:
:tada: This pull request is included in version 0.17.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.17.0).

PR #90, opened 2026-08-27
feat: un-nest sub-cards by dragging, and add columns from list view

aktasfatih:
Follow-on to #89 (merged, released as `v0.17.0`). Two cards the board owner asked for directly while reviewing that work.

## What's in here

| --- | --- | --- |
| `35e9153` | #9850 | Drag a sub-card out of its parent in list view to make it top-level again — the inverse of the nesting that shipped in #89. |
| `9fc0287` | #9853 | Add a column from list view, instead of switching back to the board first. |

## Why un-nesting is list-view-only

#89 shipped drag-to-**nest** but deliberately not the inverse, because the obvious rule ("any reorder-mode drop of a sub-card clears its parent") is silent data loss on a kanban board: a sub-card renders as an ordinary tile, so a routine "move this subtask to the next column" would break the relation with no signal and no undo.

This PR resolves that properly rather than by loosening the rule:
- **List view** gets real drag-out un-nesting — children render indented there, so dragging out of the indent is a gesture you can see yourself performing. Distinct dashed "Drop to remove from parent" affordance; reordering among siblings does not detach.
- **Kanban tiles** gain a `↳` sub-card marker, so nesting stops being invisible — but kanban drag stays non-destructive. Both #89 guard tests still pass.
- The explicit "Detach from parent" panel action remains the accessible path.

### A review pass caught three real bugs here, all fixed and covered

1. **Levels are compared as RENDERED, not stored.** A sub-card whose parent sits in another column renders flush-left — under the naive rule an ordinary reorder would have detached it. That's the same data-loss failure mode #89 rejected, hiding inside the "safe" view.
2. The group-header / column drop no longer changes a parent at all (it had no affordance suggesting it would).
3. Parent-by-drag is gated on `NEST_ENABLED`, so cross-board Views don't render dead pills.

Dropping onto a card's *own parent row* is never a detach either — that pixel boundary is shared with the first child row's top edge.

## Add column in list view

The ⋯ More → "Add column" action was never kanban-gated — it rendered in list view and **silently did nothing**, because it reveals a composer that only exists in the kanban track. Now the one action reveals and focuses whichever composer belongs to the active view. The list composer is rendered **outside the virtualiser**, so scrolling can't recycle away focus or a half-typed draft (there's a test for exactly that, across a 50-row recycle).

## Behaviour change worth knowing

`list-view.spec.js`'s "child rows are inert" spec encoded the old rule and was **rewritten** to assert the new one. That's a deliberate semantic change, not a test being loosened to pass.

## Verification

`35e9153`: 5 new e2e + 29 regression specs green, including both #89 nesting guards.
`9fc0287`: 4 new e2e (create-in-place + drag into the new column + reload; menu action focuses the list composer; focus/draft survive a virtual-list recycle; a read-only peer sees neither composer nor menu item) + 32 regression specs green.
`npm run build` clean; l10n extract/compile/lint run last in both. No `<version>` change.

### DIFF
--- a/src/components/BoardListView.vue
+++ b/src/components/BoardListView.vue
@@ -126,10 +126,13 @@ SPDX-License-Identifier: AGPL-3.0-or-later
 					</p>
 				</div>
 
-				<!-- Card row (top-level only gets DnD; children and composer are inert) -->
+				<!-- Card row. Top-level AND child rows are draggable: the list draws the
+				     indent, so dragging a sub-card out of it is a gesture the user can
+				     see themselves make (the kanban board draws no indent, so a tile
+				     drag there never touches the parent). -->
 				<div
-					v-else-if="rows[vRow.index].type === 'card' && !rows[vRow.index].isChild"
-					v-card-dnd="{ card: rows[vRow.index].card, sortMode: props.sortMode }"
+					v-else-if="rows[vRow.index].type === 'card'"
+					v-card-dnd="{ card: rows[vRow.index].card, sortMode: props.sortMode, rowParentId: rows[vRow.index].isChild ? rows[vRow.index].card.parentCardId : null }"
 					class="board-list-row-wrap"
 					:class="{
 						'board-list-row-wrap--dragging': isDraggingCard === rows[vRow.index].card.id,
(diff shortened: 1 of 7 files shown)


aktasfatih:
:tada: This pull request is included in version 0.18.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.18.0).

PR #91, opened 2026-08-27
feat: trustworthy Views, an unclipped timeline, and PR-title card linking

aktasfatih:
Sprint `sprint-2026-08-27` — **make the Views surface trustworthy, and unclip the timeline.** Nine cards, each validated against the code before implementation and each shipped with an executed, non-vacuity-proven test.

## What's in here

| --- | --- | --- |
| `d475205` | #9859 | View tiles update on edit instead of after up to 60s — `['view-cards']` was a cache island nothing invalidated. |
| `e8033f4` | #9858 | Timeline renders track, gridlines and bars for every row, not just the first screenful. |
| `5391145` | #9860 | Saved Views get a sort control, stored in the view record and applied server-side before the cap. |
| `a44cdaa` | #9846 | Bundled npm production bumps cut a patch release instead of parking on `main`. |
| `b6f395c` | #9854 | The discussion panel collapses, giving a card the full width. |
| `3c00dce` | #9855 | A PR links to its card by a `PREFIX-<seq>` reference in the PR title. |
| `d9744ec` | #9857 | Read-only members no longer see an add-card composer the server refuses. |
| `44955a1` | #9856 | "Add column" is no longer offered in views that have nowhere to put it. |
| `0e30189` | #9863 | The MCP board tool hides archived cards unless asked for them. |

## Three defects found by verification, not by the reports

Each of these was in a card as written, and would have shipped:

1. **The timeline's sticky date axis was never pinned.** The card assumed a `height: max-content` fix would *break* pinning; measurement showed `axis.top − body.top` was −1693px both before and after, identically. The axis was glued to a scroll container that never scrolls vertically. The single-scroll-container restructure fixes the clipping *and* pins both the axis and the pane header — the latter a second bug nobody had reported.

2. **A case-insensitive PR-title regex would have moved the wrong card.** On a board with prefix `KANSO`, a PR titled "Merge kanso-42 into main" matched as seq 42 and would have linked and auto-moved the unrelated card holding `board_seq` 42. Caught in review, fixed by dropping `/i`, pinned by a 5-row data provider.

3. **The collapse rules would have shipped a blank Discussion tab on mobile.** `.card-modal--discussion-collapsed .card-modal__body` (0,2,0) beats the media query's `.card-modal__body` (0,1,0) regardless of source order, so a desktop collapse followed by a narrow-viewport open would hide the pane behind the tab. Scoped with `@media not all and (max-width: 680px)` — the exact complement, so no fractional viewport width falls between the two blocks.

## Security

`3c00dce` extends the GitHub webhook's egress rule rather than working around it. The 200 response body reaches GitHub's delivery log, readable by repo admins who need not be board members, so it may name a card id only as an echo of what the sender supplied or when the card is public. A branch-derived id is an echo; a **title**-derived id is not — the sender supplied `KANSO-14`, not the internal id. Title matches therefore pass `isPublicCard()` before being named, pinned by three tests including one asserting the branch echo exemption still holds.

One residual is documented rather than silently changed: `handled: true` with `cardId: 0` still confirms that a sender-supplied sequence resolved to *something* non-public. That is the pre-existing behaviour of the issue path; returning `handled: false` would misreport a real link and move.

## Testing

Every card's spec was **executed**, and non-vacuity proven by reverting the fix and confirming the test fails — the discipline that matters, since several of these bugs are invisible to assertions that only count DOM nodes.

- PHPUnit 1488 · psalm clean · php-cs-fixer clean · vite build clean · l10n extract/compile/lint clean
- New Playwright specs: `view-live-edit`, `card-discussion-collapse` (8), `card-composer-acl` (3), plus additions to `timeline-view`, `views`, `board-rename-add-column`
- New non-Playwright suites: `tests/unit/queryKeys.test.mjs` (wired into `build-frontend` as `npm run test:unit`) and `mcp/tests/test_tools.py` (34 pytest passing)

One caveat worth knowing for future PHP work: the dev container runs `opcache.revalidate_freq=60`, which made one non-vacuity check pass against stale bytecode before it was caught.

## Notes for review

- `#9846`'s real acceptance is the **next weekly Dependabot run's commit subjects** — no CI check touches `dependabot.yml`, since it is consumed by GitHub's service. If `prefix-development` doesn't fire on grouped PRs, dev bumps also land as `fix(deps)` and cut spurious patch releases: noise, not a withheld security patch.
- `#9856` leaves one unsatisfying state: an editor on a board with **zero** columns, in timeline or swimlane mode, has no add-column affordance. Not a capability regression — the old menu item did nothing there either — and the escape hatch is one click in the Display menu. The alternative (auto-switching the view from a menu click) was rejected as its own surprise.
- No `<version>` bump anywhere; releases stay semantic-release's job.


aktasfatih:
:tada: This pull request is included in version 0.18.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.18.0).

Issue #65, opened 2026-08-21
[Bug] Recurring tasks duplicate unexpectedly and set due date to today


dbmtrde:
I've encountered a rather serious bug related to recurring tasks. For some reason, existing recurring tasks are unexpectedly duplicated.

The app seems to randomly clone or duplicate existing recurring task cards. For the newly created duplicate, the due date is automatically set to the current date (today).

**To Reproduce:**
I haven’t yet identified the exact trigger, but it occurs with existing recurring tasks on the board. It may be related to automation jobs running in the background that are executing incorrectly or being triggered multiple times. Or it could be related to a date change.

PR #96, opened 2026-08-28
feat: create and manage repeating cards from the MCP server

aktasfatih:
## What

The MCP server could read and write nearly every card property but had **no access to recurrence at all**, so an assistant could not set a card up to repeat. This adds the full rule surface, plus the one API gap that blocked using it properly.

### MCP tools

| Tool | Endpoint |
| --- | --- |
| `kanso_list_recur_rules` | `GET /boards/{id}/recur-rules` |
| `kanso_create_recur_rule` | `POST /boards/{id}/recur-rules` |
| `kanso_update_recur_rule` | `PATCH /recur-rules/{id}` |
| `kanso_delete_recur_rule` | `DELETE /recur-rules/{id}` |
| `kanso_recur_rule_create_now` | `POST /recur-rules/{id}/create-now` |

Recurrence is board automation rather than a card field, so it can't ride along in `kanso_update_card`. The tool docs spell out the actual flow — create the card, then anchor a rule on it as the template — along with the RRULE examples, CLONE vs RESET mode, and the due-date policy values an assistant has to pick from.

### API: a settable schedule timezone

`RecurrenceService::create` hard-defaulted the rule's timezone to the personal timezone of whoever created it, with no override. A rule set up for a team in another zone therefore fired at the wrong local hour, and an MCP client had no way to correct it. `POST`/`PATCH` now accept an optional IANA `timezone`:

- an unparseable id is a **400**, not a silent fallback — a schedule quietly expanded in the wrong zone fires at the wrong wall-clock hour forever;
- null/empty means "leave it alone" (a rule always expands in *some* zone, so it isn't clearable);
- a zone change counts as a schedule change and re-arms the cursor, since re-anchoring the same RRULE elsewhere shifts every future occurrence. It goes through the existing `firstFireFor`/`anchorFor` path, so the #65 and #80 protections still hold.

## Testing

- `mcp`: **46 pytest tests pass** — client-level coverage for all five endpoints (body shape, `0`/`false` surviving the None-filter, the `{"card": ...}` unwrap) plus tool-level tests, including a schema check that only `board_id` / `template_card_id` / `target_stack_id` / `rrule` are required.
- PHP: **1497 PHPUnit tests pass**, 9 of them new (explicit zone wins over the owner default, empty falls back, garbage rejected on both create and update, changed zone re-arms, unchanged zone does not).
- `psalm` clean, `php-cs-fixer` clean, l10n extract/compile produces no drift.

Not verified: a live HTTP round-trip. The local dev stack is bind-mounted to a different worktree, so it can't serve this branch. The only seam that leaves untested is the app framework binding a `?string $timezone = null` body param — which the `?string $rrule = null` param on these same two endpoints already exercises.

## Note for reviewers

CI has **no Python job**, so `mcp/tests` never gates a PR. Happy to add one in a follow-up.

### DIFF
--- a/lib/Controller/RecurRuleController.php
+++ b/lib/Controller/RecurRuleController.php
@@ -19,7 +19,7 @@ use OCP\IUserSession;
  * Recurring-card rule endpoints. Rules are board-automation config, so
  * mutations require MANAGE and listing requires READ (enforced in
  * {@see RecurrenceService}). create-now spawns one card immediately. Errors
- * map through ApiErrorTrait (bad RRULE / mode / policy → 400).
+ * map through ApiErrorTrait (bad RRULE / mode / policy / timezone → 400).
  */
 class RecurRuleController extends Controller {
 	use ApiErrorTrait;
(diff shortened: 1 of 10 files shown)


aktasfatih:
:tada: This pull request is included in version 0.19.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.19.0).

PR #97, opened 2026-08-29
fix: read-only affordances, cross-board feed cadence, Views filter correctness + parity/wire guards

aktasfatih:
Every card was re-validated against `origin/main` before execution; three of the four had drifted line refs and two had substantively wrong claims, corrected on the cards.

## What's in it

**`fix:` Hide column actions, Delete column and card drag from read-only members** (KANSO-23)
A read-only board member could see the column actions menu — including **Delete column** — an editable-looking column title, and could pick up and drag cards. Every one of those writes was already rejected server-side; they were dead affordances that read as a broken app. `BoardView` now gates all six stack callbacks on `canEditBoard` and provides a `BOARD_CAN_EDIT` inject consumed by `CardTile`, `BoardListView` and `StackColumn`. **No server-side check was loosened.**
The inject defaults to `true` on purpose: `ViewKanbanColumn` renders `CardTile` with no provider, so a `false` default would silently kill drag on cross-board Views.
Review also caught a reactivity bug: the permission was read only at mount while the grab-cursor class was reactive, so a mid-session ACL flip via the 60s refetch would desync into a dead grab handle. Both drag registrations now re-register under `watch`.

**`fix:` Refresh Views and My Work when a checklist item or the priority shortcut changes a card** (KANSO-24)
Completes #9859. The `0`–`4` priority shortcut and the three checklist mutations that change `{total,done}` reached neither the Views feed nor the My Work feeds, so a View filtered on those facets went stale for up to 60s.
Deliberately **not** routed through `usePriority`: it resolves `cardId` lazily inside `onError`/`onSettled`, so moving focus (j/k) mid-flight would roll back the wrong card. Uses the one-line `.finally()` precedent from the `d` shortcut in the same file.

**`ci:` Run the MCP server's Python test suite in CI** (KANSO-25)
`mcp/` had 46 passing tests and nothing in the repo ran them. Adds a `unit-mcp` job, plus the five keys `BoardDetail` was silently dropping (`reviewTypes`, `cardFields`, `blocksEdges`, `acl`, `subscription`) — a real MCP capability gap, not just hygiene. `extra="ignore"` deliberately left alone; the feared regression is pinned by a targeted `archived` round-trip test instead.

**`fix:` Search every readable card when filtering a View, not just the first 5000** (KANSO-21)
`ViewService::findMine()` capped at `MAX_CARDS = 5000` *before* any filter existed server-side, while the browser filtered afterwards — so on a large account a narrow filter searched an arbitrary window and `total` counted readable, not matching, cards.
New `lib/Service/ViewFilter.php` mirrors the client's 15-dimension predicate and runs **strictly after the per-board ACL masking loop and before the sort and cap**. The client predicate is kept as the render path, so the server filter is a superset guard: chip editing stays instant, and a looser PHP predicate can only ship harmless extra rows the client drops — never drop a match.
The facet-collapse blocker is solved by shipping the facet vocabulary separately: `participants` is accumulated in the existing board loop (zero extra queries), so filtering to one person no longer makes the assignee facet vanish. Also passes `:labels`, which makes the Labels facet reachable in Views at all for the first time.
**This is not a speedup** — `serialize()` still runs its enrichment queries per board. It fixes correctness and shrinks the payload.

## Verification

Every new test was executed green **and proven non-vacuous** by breaking the source and re-running:

- `column-actions-acl.spec.js` (3) — stashed the fix, rebuilt, watched it fail on each of the three assertions in turn.
- `view-checklist-live.spec.js` (3) — stashed both source files: all three failed (View row never left/re-entered the filtered set; `my-cards` hits 1, expected >1).
- `unit-mcp` — deleting a `BoardDetail` field fails the job (exit 1); renaming `uv.lock` away makes `--frozen` exit 2, proving the lock is load-bearing.
- Views filter — four independent breakages: moving the filter after the cap (4 PHPUnit failures), loosening `checklist:incomplete` in the PHP predicate, the same loosening in the JS predicate (caught by the parity fixture in the mjs runner), and removing the `participants` seed (caught by the facet e2e).

Totals: **PHPUnit 1513 tests / 912507 assertions green**; `npm run test:unit` **50/50**; new e2e specs green alongside the existing `views`, `board-filters`, `checklist`, `keyboard`, `priority`, `dnd*`, `list-view` and `swimlanes` suites; `cs-check` clean; psalm "No errors found".

The >5000-card criterion is asserted at the PHPUnit level (rows built past the cap where the only matches sit outside the capped window), not as a fake 5000-row browser test. A new ACL test proves a filter cannot surface a card from a non-readable board and that `participants` stays scoped to the readable set, so the facet vocabulary can't become a uid side channel.

## Maintainer actions

1. **`unit-mcp` is not required yet** — making it required is a branch-protection setting, not a repo file. Flip it in Settings → Branches after the first green run. The lockfile is committed so that's now safe to do.
2. `CLAUDE.md`'s "Required checks" list was deliberately left unchanged until that flip actually happens.

## Also on the board (no code here)

- **KANSO-27** — investigated whether the `E2E_ISOLATE` block in `helpers.js` defeats the `storageState` opt-out, which would have meant the anonymous public-share assertions were passing vacuously inside a required check. **Disproven** with a standalone minimal repro: `test.use` wins in Playwright 1.62.1, all 15 opt-out sites are fine. Closed with no code change.
- **KANSO-20 / KANSO-26** — the two red Dependabot PRs (#85 lockfile drift, #95 an upstream `@nextcloud/vite-config` peer pin that no in-repo change can fix) are in Dev Priority awaiting a decision; both need GitHub-side writes.

---

## Addendum — adversarial review pass (3 blocking findings fixed)

After the four cards above shipped, an adversarial review ran over the whole delta. It confirmed the highest-risk parts are sound — the filter provably sits after the per-board ACL masking loop and can never surface a non-readable card, `participants` is accumulated only from already-masked rows, field-type parity between the PHP and JS predicates is exact, `cards()` resisted every attempt to make it 500, nothing above PHP 8.0 syntax, the CSV retry is bounded to exactly once, and none of the four performance bets regress (`findMine()` adds one in-memory O(n) pass and zero queries).

It also found three blocking defects, now fixed:

**`da47cbc` `fix:` keep numeric usernames in a View's assignee and owner filters**
`participants` was built as a PHP array *key* set, and PHP coerces canonical decimal string keys to `int`. A numeric uid like `12345` (routine with LDAP employee-number provisioning) shipped as a JSON number and was then dropped by the client's `typeof u === 'string'` guard — **reintroducing the exact facet self-narrowing bug `participants` was added to prevent**. Filter to one person and the numeric-uid user vanished from the facet with no way to add them. Fixed with `array_map('strval', …)`; proven by reverting the line and watching the new test go red on `- '12345' / + 12345`.

**`147179f` `ci:` make unit-mcp assert the lockfile is in sync**
`uv run --frozen` means "don't update the lock" — it does *not* check the lock matches `pyproject.toml`. Reproduced: adding an unlocked `rich>=13` still gave `48 passed, exit 0`. So a contributor bumping a dep and forgetting `uv lock` would get green CI that tested the **old** resolution, defeating the point of committing the lock. Now `--locked`, which correctly exits 2 with "the lockfile needs to be updated".
The same commit corrects `CONTRIBUTING.md`, which had listed `unit-mcp` among the checks required to merge — live branch protection returns only the six, so it is advisory until the flip.

**`a1b21c5` `chore:` don't offer the label facet in cross-board Views**
The Views commit opportunistically passed `:labels` to turn on the label facet. But label ids collide across boards — `ViewService.php:139-144` documents this as "acceptable for chip colouring" — so with board A's label 5 = "Bug" and board B's label 5 = "Urgent", the facet showed one name while both predicates matched the bare id: **pick "Bug", get board B's "Urgent" cards**. Newly user-reachable, and wrong results rather than wrong colours, so the one line is reverted and the facet is hidden in Views again. Doing it properly needs board-qualified label identity, tracked as KANSO-29.

Full suite after the fixes: **PHPUnit 1517 / 912537 green**, `test:unit` 50/50, psalm clean, php-cs-fixer clean, 12/12 on the affected e2e specs.

### Non-blocking findings, filed as cards rather than fixed here

- **KANSO-28** — the `d` and `0`–`4` keyboard shortcuts are still ungated for read-only members, on the very surface the first commit set out to gate. Same dead-affordance class; the server correctly refuses.
- **KANSO-29** — board-qualified label identity for the Views label facet (the follow-up to the revert above).
- **KANSO-30** — parity-fixture gaps: no `now`-relative boundary cases, no sparse-row case on the JS side, and the fixture pins the two predicates but not the *wire*, so a renamed short key or a 16th dimension missing its controller param would pass green.
- **KANSO-31** — throttle the cross-board refetch on checklist edits (8 ticks in a card overlay on a View = 8 refetches of the heaviest query), and guard drag-handle teardown mid-drag.
- **KANSO-32** — the MCP container still installs unpinned; `uv.lock` pins CI only.

One review finding was **rejected**: it flagged the `E2E_ISOLATE` `storageState` override as inert, which would mean the anonymous public-share specs assert nothing in CI. That exact hypothesis was tested empirically earlier in this session and **disproven** with a live probe and a standalone repro — `test.use` wins in Playwright 1.62.1. The reviewer reached its conclusion by reading the code and noted it could not execute the suite. Executed evidence wins; details on KANSO-27.

---

## Addendum 2 — second `/pm` session (2026-08-29, later)

Eight more commits. **Every card was re-validated against the branch before execution, and grooming
caught material errors in three of the four it inherited** — two cards prescribed the wrong fix, and
two acceptance criteria would have false-passed. Details on each card.

### Shipped

**`ci:` `af25e0f` — pin setup-uv to an existing tag** (KANSO-36)
`unit-mcp` was the only red check here: `astral-sh/setup-uv@v10` does not resolve — upstream stops
publishing bare major tags after `v7`, so the v10 line exists only as `v10.0.1`. The maintainer and a
worker reached the identical fix 95 seconds apart; the duplicate was dropped and this commit kept.
Every other action pin in `ci.yml` was audited and resolves.

**`fix:` `e73af17` — the done and priority keyboard shortcuts no longer fire for read-only members** (KANSO-28)
Completes the read-only affordance work. `d` and `0`–`4` wrote with no permission check and were
advertised in the shortcuts overlay; the server correctly refused with 403 `{"error":"Access denied"}`,
so it read as a broken app. Gated on the same `canEditBoard` computed, read directly — `BoardView`
*owns* that ref and provides it, so the inject used by descendants would have been indirection for
its own sake. **Grooming caught a second live entry point the card missed:** `:on-card-focus` is also
unconditional on the swimlane path (`:411`), and the proof shows it — reverting the fix records
2 PATCHes there. Non-vacuity proven three ways.

**`fix:` `4d863df` — rapid edits in a card opened from a View refresh the feed once, not once per edit** (KANSO-31)
A card overlay on `ViewPage` keeps `view-cards` *active*, so 8 checklist ticks fired 8 refetches of
the heaviest query in the app. **The card's hedge that TanStack probably collapses these was
disproved:** `invalidateQueries` defaults to `cancelRefetch: true`, but `getViewCards` carries no
AbortSignal, so a cancelled refetch still completes server-side. Fixed at the `queryKeys.js` choke
point rather than the three checklist call sites — those are 3 of ~20 callers with the same shape.
Leading edge fires immediately so the View still repaints; `view-checklist-live.spec.js` passes
unchanged, which is the proof the leading edge is intact.

**`chore:` `e2174af` — clear the dragging flag when a permission flip tears the drag handle down** (KANSO-31)
Two lines. **The card prescribed porting `BoardListView`'s `pendingRebinds` deferral; that was wrong**
— it guards *drop targets*, and this path touches only the draggable. One of the card's two stated
triggers is also unreachable (a falsy `boardData` unmounts the tile outright). The comment records
why the deferral is deliberately not copied.

**`test:` `bc2b76a` + `8903644` — close the parity-fixture gaps** (KANSO-30)
The golden fixture pinned the predicates but not the boundaries, not sparse rows, and **not the wire**.
The last is the one that mattered: a symmetric short-key rename left both unit runners green while the
client sent a key the server never read — reverting the whole point of server-side View filtering,
invisibly, because the client still re-filters what it receives. Fixed by deleting the hand-copied key
map (parsing the real one instead) and adding reflection guards. Red output was produced for every
guard, and the new sparse row was **proven to be the sole discriminator** — delete it and the same
defect passes 45/45.

**`fix:` `41a9637` — the error banner clears itself once the next action succeeds** (KANSO-37)
The banner was only ever cleared by its ×, so a stale failure survived a later *successful* action.
`moveError` had the same defect and is materially worse — it wins the `moveError || shortcutError` OR
and masks every later shortcut error. Validation grew the scope by two lines: the ref has **four**
writers, and clearing on two of four would have been less coherent than the bug.

### Adversarial review of the above, and the fixes it forced

A review pass over `af25e0f..41a9637` **verified the delta is safe on the source side** — no
production defect — and confirmed by mutation testing that the new tests genuinely fail against
broken source. It also found one real gap and several small ones, all now fixed:

**`test:` `a1710fb` — close the wire guard's blind half and pin the week window's width**
The headline new guard was **half-blind and its docblock overclaimed.** `ViewController::cards()`
binds filter params in *two* places — the parameter list and a hand-written `$query` literal — and the
guard only checked the first. Deleting `'fsc' => $fsc,` from the literal left all ten tests green
while the server silently stopped constraining `subcard`: exactly the failure mode the docblock
claimed to catch, and the half more likely to drift. Also: the fixture pinned the boundary
*operators* but not the window *width* (`WEEK_MS` 7d→14d survived undetected, and it is duplicated
across PHP and JS), fixed with one card at `now + 8d`. Two factually wrong comments corrected — the
CSV retry rationale described an unreachable scenario, though its assertion count was right.

**`fix:` `5f1e776` — a failed card move keeps its explanation until you act again**
Three real behaviours. (1) The move-queue race flagged when `41a9637` shipped is **confirmed, not
refuted**: moves are serialised, so a queued drag B that succeeds wiped failed drag A's banner while
A's rollback was still on screen. The review found a simpler fix than the containment considered and
rejected — clear on *enqueue*, not on success, so a new gesture retires the banner but a queued
success cannot. (2) A backwards clock step (NTP correction, VM resume) made the new debounce's
`elapsed` negative and could **wedge View-feed invalidation off entirely**; now clamped. (3) The
`.catch` had come to chain after the `.then`, so a throw from the invalidation would paint a failure
over a write that actually succeeded.

### Verification

PHPUnit on **8.2 and 8.3**: 1518 tests / **912,584 assertions** green. `npm run test:unit`: **61/61**
(50 at session start). psalm "No errors found!"; php-cs-fixer "0 of 372 files"; `npm run build` OK;
l10n extract/compile/lint clean. e2e specs run at `--retries=0` throughout.

**Every new guard has pasted red output against the defect it targets** — the deleted `$query` entry,
`WEEK_MS` on both sides, the unclamped clock delay, and the restored clear-on-success each turn a
runner red. The dev container's bind mount was re-verified with `docker inspect` before any e2e was
trusted; it had been pointing at a pruned sibling worktree, which would have silently tested other code.

### Deliberately left open

- **`getViewCards` still has no AbortSignal** (KANSO-39, parked) — the debounce bounds how often a
  cancelled-but-completed refetch happens; it does not remove it. The remaining win depends on whether
  PHP-FPM actually stops work on client abort, which has not been measured.
- **`BoardView` still renders `moveError || shortcutError`**, so clearing one can re-expose the other.
  Collapsing them into a single ref is the clean fix and stays deliberately out of scope.
- **Six hand-maintained parallel dimension lists in `useBoardFilters.js`** (KANSO-41, parked) — this
  work pinned one of them.
- **KANSO-29** (board-qualified label identity for the Views label facet) is in Dev Priority awaiting
  a decision. Validation settled the design and found the card's own suggested cheaper alternative
  does not actually fix anything; what remains is a priority call.

### Maintainer actions

Unchanged from Addendum 1, plus: `unit-mcp` should now actually run — it could not before, so its
first green result on this PR is the first real signal from that job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,41 @@ jobs:
       - name: Run unit tests
         run: vendor/bin/phpunit -c phpunit.xml
 
+  # The in-repo Python MCP server (mcp/) ships its own pytest suite, and until
+  # now nothing in the repo ran it — mcp/Dockerfile only `pip install .`s the
+  # package, so every MCP change was verified purely by whoever happened to run
+  # pytest locally.
+  #
+  # Why its own job rather than a step bolted onto build-frontend (as the
+  # info.xml and l10n guards were): those are Node-toolchain checks with a
+  # natural host job. Python has no such host here, and wiring uv into the Node
+  # job purely to keep the job count at six is a contortion, not a saving.
+  # Likewise no `paths:` filter — paths-scoping a *required* check drags in
+  # required-check semantics (a skipped job never reports) for a job that costs
+  # seconds. The project's aversion was to CI cost and flakiness, not to a
+  # number.
+  #
+  # `--frozen` + the committed mcp/uv.lock make the run hermetic: pyproject
+  # pins only floors, so without the lock every run would resolve fresh from
+  # PyPI and an upstream release could redden an unrelated PHP-only PR.
+  unit-mcp:
+    runs-on: kanso-runners
+    defaults:
+      run:
+        working-directory: mcp
+    steps:
+      - uses: actions/checkout@v7
+      - name: Install uv
+        uses: astral-sh/setup-uv@v10
+        with:
+          # Don't depend on whatever Python the runner image happens to ship —
+          # uv fetches its own managed interpreter (pyproject needs >=3.11).
+          python-version: '3.12'
+          enable-cache: true
+          cache-dependency-glob: mcp/uv.lock
+      - name: Run MCP server unit tests
+        run: uv run --frozen --extra dev pytest
+
   build-frontend:
     runs-on: kanso-runners
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,7 +104,11 @@ jobs:
     steps:
       - uses: actions/checkout@v7
       - name: Install uv
-        uses: astral-sh/setup-uv@v10
+        # setup-uv stopped publishing floating major tags after v7, so this is
+        # pinned to an exact release deliberately and needs a manual bump.
+        # Do not "simplify" it to @v10 — that tag does not exist and the job
+        # then fails to start with "unable to find version".
+        uses: astral-sh/setup-uv@v10.0.1
         with:
           # Don't depend on whatever Python the runner image happens to ship —
           # uv fetches its own managed interpreter (pyproject needs >=3.11).
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,9 +87,15 @@ jobs:
   # seconds. The project's aversion was to CI cost and flakiness, not to a
   # number.
   #
-  # `--frozen` + the committed mcp/uv.lock make the run hermetic: pyproject
+  # `--locked` + the committed mcp/uv.lock make the run hermetic: pyproject
   # pins only floors, so without the lock every run would resolve fresh from
   # PyPI and an upstream release could redden an unrelated PHP-only PR.
+  #
+  # `--locked`, NOT `--frozen`. Both install strictly from the lock, but only
+  # `--locked` first ASSERTS the lock is in sync with pyproject and fails if it
+  # is not. Under `--frozen` a contributor who bumps a dependency and forgets
+  # `uv lock` gets a green run that quietly tested the OLD resolution — which
+  # defeats the point of committing the lockfile at all.
   unit-mcp:
     runs-on: kanso-runners
     defaults:
@@ -106,7 +112,7 @@ jobs:
           enable-cache: true
           cache-dependency-glob: mcp/uv.lock
       - name: Run MCP server unit tests
-        run: uv run --frozen --extra dev pytest
+        run: uv run --locked --extra dev pytest
 
   build-frontend:
     runs-on: kanso-runners
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -31,7 +31,8 @@ php-cs-fixer). Before opening a PR:
 
 `main` is a protected branch: every change lands through a pull request and can
 only be merged once **all** CI checks pass — `cs-check`, `psalm`, `unit-php` (on
-PHP 8.2 and 8.3), `build-frontend`, and the full `e2e` suite. Open your PR
+PHP 8.2 and 8.3), `unit-mcp` (the `mcp/` Python server's pytest suite),
+`build-frontend`, and the full `e2e` suite. Open your PR
 against `main`; CI runs automatically on push.
 
 ## Translations
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,10 +30,15 @@ php-cs-fixer). Before opening a PR:
   the `kanso_changes` delta log, ETags) — don't regress these.
 
 `main` is a protected branch: every change lands through a pull request and can
-only be merged once **all** CI checks pass — `cs-check`, `psalm`, `unit-php` (on
-PHP 8.2 and 8.3), `unit-mcp` (the `mcp/` Python server's pytest suite),
-`build-frontend`, and the full `e2e` suite. Open your PR
-against `main`; CI runs automatically on push.
+only be merged once **all** required CI checks pass — `cs-check`, `psalm`,
+`unit-php` (on PHP 8.2 and 8.3), `build-frontend`, and the full `e2e` suite.
+Open your PR against `main`; CI runs automatically on push.
+
+`unit-mcp` (the `mcp/` Python server's pytest suite) also runs on every PR, but
+it is **advisory for now** — it is not yet in the branch-protection required
+list, so a red `unit-mcp` will not block a merge. Adding it there is a
+maintainer action in the repository settings that has not happened yet. Please
+treat it as required anyway: don't merge on a red `unit-mcp`.
 
 ## Translations
 
--- a/REUSE.toml
+++ b/REUSE.toml
@@ -11,6 +11,7 @@ path = [
     "package.json",
     "package-lock.json",
     "tests/psalm-baseline.xml",
+    "tests/fixtures/**",
 ]
 precedence = "aggregate"
 SPDX-FileCopyrightText = "2026 Fatih AKTAS <akfatih2@gmail.com>"
--- a/lib/Controller/ViewController.php
+++ b/lib/Controller/ViewController.php
@@ -9,6 +9,7 @@ namespace OCA\Kanso\Controller;
 
 use OCA\Kanso\Service\InvalidInputException;
 use OCA\Kanso\Service\NotPermittedException;
+use OCA\Kanso\Service\ViewFilter;
 use OCA\Kanso\Service\ViewService;
 use OCP\AppFramework\Controller;
 use OCP\AppFramework\Http\Attribute\NoAdminRequired;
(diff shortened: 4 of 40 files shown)


aktasfatih:
:tada: This pull request is included in version 0.19.1 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.19.1).

PR #98, opened 2026-08-31
fix: recurrence, webhook and description-conflict correctness, plus CI and dev-stack reliability

aktasfatih:
**14 commits · 32 files · +1916/−178 · no version bump · one unbumped migration.**

## What ships

**Recurrence**
- A repeat set to stop after N times now actually stops. `COUNT` was evaluated against a DTSTART that reset-mode legitimately moves, so the window restarted on every spawn. Capped from the rule's existing `occurrences_spawned` tally — no migration needed. Editing the schedule now resets that tally, so "end after 4 more" gives you 4.
- Pointing a repeat at another card reschedules it from that card. Previously it kept firing on the old card's anchor.
- The docs claimed a schedule is anchored at the rule's creation time. It isn't — it's the template card's start date, else its due date, else creation time. Corrected at five sites, including two the card missed, and pinned by the repo's first test that asserts occurrence dates for a `BY*` rule part.
- An unreadable repeat schedule in an imported board is skipped instead of stored, and no longer crashes the import.

**GitHub webhook**
- The setup panel and README now say the Content type must be `application/json`. With the form-encoded option the body still passes the HMAC check, so a misconfigured webhook returned a green 200 and silently did nothing — indistinguishable from a ping in GitHub's delivery log.
- A body that isn't JSON now returns 400 with a hint instead of that silent 200, and a *downstream* failure no longer misreports itself as a content-type problem.

**Card descriptions**
- A description edit can no longer be lost to a simultaneous save. Adds a `description_revision` counter (migration, unbumped) and claims it with a conditional UPDATE inside the existing change-log transaction, so a lost race writes no card row, no change row and no realtime push.
- The Phase 1 `baseLastModified` guard is deliberately **kept**, gated on the new parameter being absent — it shipped in v0.19.1, and removing it would silently turn a public API guard into a no-op for third-party clients.

**CI and dev stack**
- PHP jobs install `libsodium23` if it's missing, so a runner without it stops failing `php` with exit 127 before any test runs, plus a `php -v` tripwire that names the failure.
- deck and contacts are provisioned from one script shared by `dev/setup.sh` and CI, instead of being duplicated in the workflow file where only one copy was maintained. Two e2e specs were failing locally purely for that reason.
- An offline or appstore-unreachable dev boot now warns and continues instead of dying.

## Review

Each card got an adversarial review during implementation. The **whole delta** was then reviewed separately, which caught things per-card review structurally couldn't — most importantly a crash introduced by the import guard itself: sabre's `RRuleIterator::$frequency` is a typed uninitialised property with no FREQ-required check, so `''` / `COUNT=3` construct cleanly and throw `\Error` past every `catch (\Exception)`. Older exports contain empty rrules, so that would have turned a working import into a 500. Six findings total, all closed here with mutation evidence.

## Verification

- PHPUnit **1552 tests / 912,704 assertions** green on PHP **8.2 and 8.3** (1518 at branch point).
- `psalm` exit 0 · `php-cs-fixer` 0 of 376 · l10n extract/compile idempotent.
- New tests were **mutation-tested**: each was re-run with its fix reverted to prove it actually fails. Notably the description-conflict e2e fails 3 of 4 with the CAS removed while PHPUnit stays fully green — that suite mocks `CardMapper`, so it structurally cannot prove a concurrency property.
- The migration was applied against a live NC 34 + postgres stack and rolled back.
- `appinfo/info.xml` and `package.json` versions untouched.

**Not verified locally:** `e2e` and `build-frontend` — CI is the first real run of both.

## Two things worth a look before merging

1. **A deliberate behaviour change beyond the bug.** A manual "create now" now spends a `COUNT` slot in clone mode too, where previously it didn't shorten the series. Consistent semantics and test-pinned, but clone mode wasn't broken. Reverting is a small exemption on `createNow`.
2. **The `COUNT` tally reset is unconditional on a schedule edit**, so lowering `COUNT=10` to `COUNT=3` mid-series restarts the count rather than ending it three cards from the original start. Called out in the commit body; the other reading is defensible.

Please **merge with a merge commit** — squashing would collapse the typed commits and lose the release.


aktasfatih:
:tada: This pull request is included in version 0.20.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.20.0).

Issue #46, opened 2026-08-19
Feature Request: Internationalization / Multi-language support


dbmtrde:
Hi @aktasfatih,

Thanks again for the great app!

It would be great to have multi-language support so Kanso can follow the user's Nextcloud language settings.

I'd gladly help out with the German translation if you add support for it.

PR #100, opened 2026-09-01
feat: archived-card handling, card-view navigation, My Tasks, and a review-gate fix

aktasfatih:
**Every card here traces to a requirement the product owner stated**, not one inferred from a title — see "On intent" at the bottom, which is the main process note of this session.

## What's in it

**Archived cards**
- **Views no longer show archived cards**, with an opt-in `archived` filter facet (`null | include | only`) reusing the existing single-select facet machinery — no new settings surface. Exclusion happens *before* the `MAX_CARDS` cap, so `total` is correct too.
- **The card's ⋯ menu has an "Archived" checkbox** instead of a verb that flipped between "Archive" and "Unarchive". One control that shows state and changes it.

**Card view**
- **The breadcrumb shows the card's column** (`Board › Column › [status chip]`), on every board. The chip now always shows *status*, and its popover always lists columns with the statuses beneath — previously, a board with no column roles never displayed which column a card was in, and role-less is the **default** for every fresh and imported board.
- **"Find the card on board"** — from an open card, scrolls the board to it and rings it. Drives the virtualizer to the card's index (a plain `scrollIntoView` can't work: the row isn't mounted), expands a collapsed column first, and tells you when a filter is hiding the card rather than silently scrolling nowhere.

**My Tasks**
- **Recently done tasks**, in a collapsed section that fetches only when expanded — bounded by a 14-day window *and* a 50-row cap. The default page load is unchanged and issues no extra query.
- **Four defects fixed**: an invisible focus ring on rows that are explicitly `role="button"` (WCAG 2.4.7), Space not activating them, a filtered-empty state that claimed you had no tasks *anywhere*, and a silent 200-row cap that also froze the nav badge at a wrong "200".

**Correctness / security**
- **A card with unapproved reviews can no longer be marked done by any route.** The gate previously lived only in `move()` and keyed on the *source* role, so `{done:true}` from the bulk bar or the `d` shortcut skipped it entirely, `maybeCompleteParent()` was a fourth ungated writer, and **Review → any role-less column → Done** laundered it in two ordinary drags. `done` is now a thin alias for `status`, and the gate keys on the target being Done plus the card's own review state.
- **Exports include attachments** — and so do **scheduled backups**, which reused the same producer and had been silently dropping every file. Envelope v3 is a streamed zip; v2 JSON still imports.
- **`"5 project"` → `"5 projects"`** — counted nouns moved to plural msgids across the frontend, with the recurrence phrases restructured so the number and its unit can agree in inflected languages.

**Tooling**
- **`kanso_search_cards`** MCP tool, wrapping the existing `/api/search`. ACL is enforced server-side; the tool passes `board_id` as a filter and never fans out.
- **CI's l10n guard now runs `sync`.** It previously ran extract + compile only, so a new string landed in the template, compiled to nothing, and every catalogue silently fell back to English with `git diff --exit-code` clean. That had already happened to 11 strings on this branch.

## Verification

Every new test was **proven non-vacuous** — the source it covers was broken, the test confirmed red, the output recorded, then restored. That discipline earned its keep repeatedly here: it caught a `startedAt` assertion that would have been vacuous at 0→0, and an MCP denial test whose mutation revealed that dropping the board scope *silently returned other boards' cards*.

At the branch tip: `test:unit` 89/89 · `test:l10n` 19/19 · `l10n:lint` clean across all ten catalogues (1,225/1,225 each, zero empty) · `l10n:check` (the new guard) passes · `views.spec.js` 9/9 · psalm and php-cs-fixer clean · no version bump, no CHANGELOG edit.

**One test was fixed, not written, in this branch.** `views.spec.js:360` was failing deterministically from the commit that introduced it — it drove the panel's global "Clear filters" believing it reset one facet, which also wiped the View's saved label filter (a dimension `ViewPage` deliberately doesn't expose), widening the list until virtualization pushed the asserted rows out of the window. It now resets the facet with its own control. Assertions unchanged. The underlying product defect is filed separately.

## Things to know before merging

- **`BackupService` now writes `.zip`, not `.json`.** Retention prunes both so old backups still age out, but external tooling globbing `kanso-board-*.json` will stop matching.
- **Bulk *Mark done* and the `d` shortcut are now column-aware.** On a board with a Done column they move the card there rather than only stamping the timestamp — the intended consequence of collapsing the write paths onto one gated path, and it matches the card-detail button, but it is a visible change to two controls.
- **A v3 `.zip` cannot be fed back to the import endpoint yet.** Import-side attachment restoration is deliberately deferred until someone asks for it — it would be the app's first attacker-controlled multi-file write path, and there's currently no zip reading, no quota accounting and no rate limiting in `lib/`.

## On intent

Six of these cards arrived as one-line titles with no description. I reconstructed the intent from the code and was **wrong every time I tried** — "statuses are confusing" was about not being able to change a column, not the status model; "redesign My Tasks" meant "show me recently done tasks"; "find card in board button" meant the *reverse* of search. Each was corrected by the owner before implementation, and each card now records the requirement in their words. The two I escalated rather than inferred cost a round-trip; the two I inferred are the two I'd have shipped wrong.

Refs #46


aktasfatih:
:tada: This pull request is included in version 0.22.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.22.0).

PR #102, opened 2026-09-02
fix: stop silent data loss across filters, cards, recurrence and archived boards

aktasfatih:
**Milestone: stop the app silently discarding things the user did not ask it to discard.**

Seven cards, seven commits. Six of the seven are the same shape of bug — a surface that quietly destroys or leaks state the user never chose — which is why they shipped together.

### Owner-filed

- **Archived boards no longer leak into any cross-board list** (`d442d4f`). v0.22.0 filtered archived *cards*; archived **boards** still showed everywhere. Root cause was one level down — `BoardMapper::findAllForUser()` filters only `deleted_at` — so the leak hit **eight** surfaces: Views, My Tasks (both feeds), Inbox, My Reviews, Search, My Steps, Projects, board folders. New sibling `BoardService::findAllActive()`; `findAll()`/`findAllWithStats()` untouched so the boards page's **Archived tab still works**. Exclusion is unconditional, deliberately not wired to the archived-cards facet.
- **Leaving a card with unsaved text now asks first** (`633c991`). No guard of any kind existed. Covers title, description, new comment, reply and inline comment edit, on X / backdrop / Escape / tab-close. Also fixes a worse adjacent bug: `<router-view>` is unkeyed, so a half-typed reply **leaked onto the next card you opened**.

### Found by validating those cards

- **Clear filters no longer wipes a filter you cannot see** (`05e1e02`). On a View it destroyed the View's own saved label filter — invisible, unrecoverable without a reload, and it widened the feed to the whole cross-board set. `clearAll()` now iterates `visibleDimensions`: a control may only clear what it renders.
- **The recurrence editor stops rewriting schedules it cannot model** (`f800a49`). An API/MCP-authored `FREQ=MONTHLY;BYDAY=1MO` became a bare `FREQ=MONTHLY` on any save. A review pass inside the card found a **second live half**: the server compares rrule *strings*, so even an equivalent-but-normalized rewrite zeroed `occurrences_spawned` — an "ends after 5" rule silently became 5 more. Custom schedules are now read-only and `rrule` is omitted from the PATCH.
- **Columns and checklist items keep a stable order between reloads** (`6729445`). Down-scoped from the filed card: the ~150-line unique-index + retry + 409 treatment was rejected as introducing a new user-visible failure mode to fix an order flicker. Two `addOrderBy('id')` calls make a duplicate sort key harmless instead of merely rarer.
- **The command palette is reachable on touch** (`337409c`). It opened only via Ctrl/Cmd+K, so on a phone the feature was unreachable — and it is advertised in the App Store listing. One item in the existing More menu; no header pressure, no new string.
- **A superseded View feed refetch is aborted in the browser** (`7c65a20`). TanStack already hands `queryFn` a signal; it was being dropped. Typed `perf`, not `fix` — the saving is client-side only, and the commit says so: the server finishes the request either way.

### Verification

Every new test proven **non-vacuous by mutation** — including the negative controls (closing a clean card must never prompt; a board where labels *are* visible must still clear them). Independently re-run before this push, not taken from worker reports:

- PHPUnit **1490 tests**, no failures (3 pre-existing PHP 8.5 deprecations)
- `test:unit` **100/100**; psalm clean; php-cs-fixer 0 of 383
- Playwright **22 passed** across the four new specs + `views.spec.js`
- `l10n:check` regenerates to a clean tree

`ImportServiceTest` / `RecurrenceServiceTest` were excluded locally — they hang under the local runner (tracked as #10093). CI runs them.

### Worth knowing before release

- `views.spec.js` previously drove the global "Clear filters" mid-test, which silently destroyed the View's identity — every assertion after it was running against an unbounded feed. Fixed.
- A new `ArchitectureTest` ratchet fails the build if a future cross-board feed calls `boardService->findAll(`.
- Browser **Back** is deliberately not covered by the unsaved-changes guard: it was implemented, measured, and removed because cancelling threw the user out of the app entirely under hash history. Filed separately.

Follow-ups filed rather than folded in: due reminders firing for archived boards, the hash-history back-navigation defect, a sub-day recurrence offset rounding to zero, and a URL-set hidden filter that can no longer be cleared.

### DIFF
--- a/lib/Db/ChecklistItemMapper.php
+++ b/lib/Db/ChecklistItemMapper.php
@@ -46,6 +46,13 @@ class ChecklistItemMapper extends QBMapper {
 	/**
 	 * All items of a card, in display order.
 	 *
+	 * Checklist items carry no unique index on their sort key and
+	 * {@see \OCA\Kanso\Service\SortKeyService::between()} is deterministic, so two
+	 * concurrent moves into the same gap can derive the SAME key. `sort_key` alone
+	 * would then leave the tied rows in whatever order the DB returns, which can
+	 * flip between reloads; the `id` tiebreaker (same idiom as
+	 * {@see self::findOpenAssignedInBoards()}) keeps that order stable.
+	 *
 	 * @return ChecklistItem[]
 	 * @throws Exception
 	 */
@@ -54,7 +61,8 @@ class ChecklistItemMapper extends QBMapper {
 		$qb->select('*')
 			->from($this->getTableName())
 			->where($qb->expr()->eq('card_id', $qb->createNamedParameter($cardId, IQueryBuilder::PARAM_INT)))
-			->orderBy('sort_key', 'ASC');
+			->orderBy('sort_key', 'ASC')
+			->addOrderBy('id', 'ASC');
 
 		return $this->findEntities($qb);
 	}
--- a/lib/Db/StackMapper.php
+++ b/lib/Db/StackMapper.php
@@ -41,6 +41,13 @@ class StackMapper extends QBMapper {
 	/**
 	 * All non-deleted stacks of a board in display order.
 	 *
+	 * Unlike cards, stacks carry no unique index on their sort key, and
+	 * {@see \OCA\Kanso\Service\SortKeyService::between()} is deterministic, so two
+	 * concurrent moves into the same gap can derive the SAME key (documented as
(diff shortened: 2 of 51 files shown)


aktasfatih:
:tada: This pull request is included in version 0.22.1 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.22.1).

PR #105, opened 2026-09-03
fix: reject a hanging repeat rule and show exact activity timestamps

### LINKED ISSUE #103: [Feature Request] Add an editable "Completed on" (Date & Time) field to tasks
Hi @aktasfatih,

I have a suggestion for an enhancement regarding task dates.

**Current state:**
Currently, task cards have options for a "Start Date" and a "Due Date", which is perfect for planning. However, there is no explicit visual record of when exactly a task was actually finished.

**Proposed Feature:**
It would be incredibly useful to have a "Completed on" (or "Done at") timestamp (Date and Time) displayed on the task.

**Implementation Idea:**
Automatic capture: The app could automatically set this field to the current date and time the moment a task is marked as "Completed" (or moved to a "Done" column).

**Manual correction:** 
It is crucial that this field remains manually editable. In the real world, you might finish a task on Friday but forget to click "Done" until Monday. Users need the ability to manually adjust the "Completed on" date to reflect reality.

**Why this is important:**
Tracking & Reviews: It makes it much easier to review past work accurately.
Documentation: It provides a clear audit trail for completed work.

What do you think about adding this to the card details? Thanks!

aktasfatih:
Started as the follow-up to #103; a validation pass alongside it turned up a **live availability bug in the released app**, so that is the headline.

## 1. A repeat rule with no `FREQ` pinned a PHP worker forever (`e5ccb39`)

**This is shipped, and reachable by any user who can manage a board.** `RecurRuleController::create()` declares `string $rrule = ''`, so a request that merely **omits** the field reaches the spin. Each hit pins one PHP-FPM worker until the pool is exhausted. On the cron path there is no `max_execution_time` at all.

Proved against **Nextcloud's own bundled sabre**, not the test `vendor/`:

```
sabre file: /var/www/html/3rdparty/sabre/vobject/lib/Recur/RRuleIterator.php
   before:  constructed … exit=124   (killed by timeout 8; never returns)
   after:   empty / whitespace / COUNT=3 / INTERVAL=2  ->  REJECTED in <1ms
            FREQ=DAILY                                 ->  RETURNED-OK
```
Real API calls agree: before, omitted-`rrule` / `''` / `INTERVAL=2` all hung at 25.01s (`curl_rc=28`) on create *and* update; after, every shape is a **400 in ~0.6s**. The cron path was proved separately with a rule seeded *before* the fix, bootstrapping real Nextcloud under the CLI SAPI: `runDueRules()` returns in 0.062s.

**Why review missed it.** `RecurrenceService.php:129` already had a `catch (\Throwable)` written for exactly this, and its comment explains sabre raises an `\Error` on a *"typed-but-uninitialized property"*. That is true of sabre **5.x**, which `composer.lock` pinned and the suite ran. Nextcloud ships **4.x**, where `$frequency` is untyped and defaults to `null`, and `fastForward()` has no iteration cap — **there is no `\Error` to catch**. Kanso ships no `vendor/` (`composer.json` has no `require` section; the release script excludes it), so production always resolves the class from `3rdparty`.

The fix is `assertHasFrequency()`, called *before* the iterator is constructed — the sole `new RRuleIterator` site, so all four callers are covered at once.

**A second defect fell out of the same baseline:** `COUNT=3` did not hang — it returned **200 and persisted a permanently-dead rule** (`nextOccurrenceAt=0`). Now a 400.

**The vacuity is closed, not worked around.** `sabre/vobject` is pinned to `^4.2` (lock: 4.6.1) so the suite exercises what actually ships. The mutation proof isolates it exactly — identical guard-less code:

```
under sabre 4.6.1 (production-matching):  PHPUNIT_EXIT=124  — hangs, no test completes
under sabre 5.0.0 (what the suite ran):   PHPUNIT_EXIT=0    — OK (5 tests, 5 assertions)
```

## 2. Activity entries now carry the exact date and time (`1a2bdfb`, `9714d87`) — closes #103

#103 asked for an editable "Completed on" field. That was declined in the thread (start/end semantics break down with recurring cards); the agreed resolution was to make the activity log exact, and the reporter accepted.

Relative label stays primary, exact stamp always visible beside it — deliberately **not** hover-only, since #10066 just fixed that class of defect for touch users. Today's rows collapse to the clock; older rows carry the date. Each is a real `<time datetime>` with a full `title`.

Two things worth noting. The separator began as a CSS `::before` and a live DOM dump showed copy-paste jamming the labels into `"just now09:31 PM"` — it is a real text node now, pinned by an assertion. And the first pass rendered `09:44 PM`: `Intl` only takes the locale's own time pattern when hour and minute are requested at the **same width**, so the naive `hour:'numeric', minute:'2-digit'` fix would have rewritten en-GB/de-DE/fr-FR `09:04` → `9:04`, fixing one locale by damaging several. Shipped as `numeric`/`numeric`, verified across 50 locales, with tests that run the helper in a child `node` under forced `LANG` — the helpers format with locale `undefined`, so a locale bug is otherwise invisible to a suite running in one locale.

## 3. Every CI job now has a runtime cap (`d998706`)

Consequence of the pin above: a future regression here fails by **hanging**, and `cs-check`, `psalm`, `unit-php`, `unit-mcp` and `build-frontend` had no `timeout-minutes` — they would burn to GitHub's 6h default. Caps sized from observed durations across 7 recent green runs with 2–3× headroom. All 7 jobs are now capped.

## Verification

Re-run independently before pushing, not taken from worker reports: **PHPUnit 1664 / 918,492 assertions** (3 pre-existing PHP 8.5 deprecations), `test:unit` **111/111**, psalm clean, php-cs-fixer 0 of 383, `activity` + `card-time-tracking` e2e 10 passed, recurrence e2e 21 passed, `l10n:check` no drift, and the end-to-end sabre proof above. Every new test mutation-proved.

Closes #103

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,10 +26,25 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Every job below carries a `timeout-minutes`. GitHub's default is SIX HOURS,
+  # which is not a safety net — a job that hangs instead of failing burns six
+  # hours of runner time per matrix leg before anyone finds out. That stopped
+  # being hypothetical once the test suite started pinning dependency versions
+  # to match what Nextcloud ships: a regression in that area surfaces as a
+  # request that never returns, not as a failed assertion.
+  #
+  # The values are sized off the observed maxima of recent green runs, with
+  # roughly 2-3x headroom, because these are self-hosted runners and a job that
+  # normally takes 3 minutes has been seen to take 13 under contention. They are
+  # a runaway guard, not a performance budget — a slow but healthy run must
+  # still pass. Re-measure before tightening any of them.
+
   # php-cs-fixer and psalm are PHP-version-independent (they analyse source,
   # not runtime), so pin them to the highest supported PHP (8.3) — one job each.
   cs-check:
     runs-on: kanso-runners
+    # Observed 3-11m across recent green runs.
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v7
       # The ARC runner image intermittently lacks libsodium.so.23, which the
@@ -104,6 +119,8 @@ jobs:
 
   psalm:
     runs-on: kanso-runners
+    # Observed 2-13m across recent green runs.
+    timeout-minutes: 30
     steps:
       - uses: actions/checkout@v7
       # See cs-check above for why this step exists and why it is not `grep -q`.
@@ -154,6 +171,9 @@ jobs:
   # range (8.2 for NC 30–33, 8.3 for NC 34) to catch version-specific breakage.
   unit-php:
     runs-on: kanso-runners
+    # One value covers both legs, so it is sized for the slow one: 8.2 has been
+    # observed at 2-6m but 8.3 at 3-21m.
+    timeout-minutes: 45
     strategy:
       fail-fast: false
       matrix:
@@ -229,6 +249,8 @@ jobs:
   # defeats the point of committing the lockfile at all.
   unit-mcp:
     runs-on: kanso-runners
+    # Observed under a minute; the cap is really for a stalled dependency fetch.
+    timeout-minutes: 15
     defaults:
       run:
         working-directory: mcp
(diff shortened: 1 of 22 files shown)


aktasfatih:
:tada: This pull request is included in version 0.23.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.23.0).

PR #106, opened 2026-09-03
fix: require Nextcloud 32 or newer, the oldest version Kanso actually runs on

aktasfatih:
Kanso declared support for Nextcloud 30 and **could not run there**. Two `IQueryBuilder` APIs it calls don't exist on NC 30:

- `PARAM_DATETIME_MUTABLE` — `CardMapper.php:1058/1391/1501`, reached from `BoardService::findAllWithStats()`, the boards-list payload. That is *opening the app*.
- `forUpdate()` — `CardMapper.php:381`, reached from `CardService::rebalanceStack()` and `CsvImportService`.

## The floor is now 32, and that is a deliberate choice

Per [Nextcloud's maintenance schedule](https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule): NC 30 went EOL 2025-10, NC 31 in 2026-02. So the breakage only ever affected people on a Nextcloud upstream stopped patching 7–11 months ago.

**32 rather than 33**, even though 32 reaches EOL on 2026-09-30: cutting to 33 today would make Kanso uninstallable for people on a version that is *still supported*. 32 is the oldest release upstream supports **and** Kanso actually runs on. Raising it again after the 30th is a one-liner.

Verified `occ app:enable kanso` is now refused on NC 30 — the floor works as an install gate.

## The real fix is the third commit-worth of change: CI now runs app code

`install-matrix` has been booting NC 30 on every PR for months and reporting green, because `dev/smoke.sh` only counted database tables. It never executed a line of Kanso.

It now creates a board, stack and overdue card over the HTTP API, then exercises both APIs:
- `GET /api/boards` → `findAllWithStats()` → `PARAM_DATETIME_MUTABLE`
- `occ kanso:rebalance --board <id>` → `findByStackForUpdate()` → `forUpdate()`

**Proven non-vacuous by running it against real containers** — this is the whole point, since a smoke check that passes on a broken platform is worse than none:

```
NC 30 →  kanso_* tables present: 34
         ok: kanso_boards / kanso_stacks / kanso_cards / kanso_changes   <- old smoke.sh stopped here, GREEN
         schema OK — now exercising app code
         SMOKE FAIL: GET /api/boards → HTTP 500     EXIT=1
```
Server log: `Undefined constant IQueryBuilder::PARAM_DATETIME_MUTABLE in CardMapper.php line 1391`.

```
NC 32 →  ok: boards list with stats (PARAM_DATETIME_MUTABLE)
         ok: occ kanso:rebalance (forUpdate)
         SMOKE OK: NC=32 DB=postgres     EXIT=0
```
Also green on NC 32 + sqlite.

## Incidental find, filed separately

`occ kanso:rebalance` **fatals on every SQLite Nextcloud, all versions** — `Operation 'FOR UPDATE' is not supported by platform`. The documented recovery for sort-key overflow is unusable on SQLite installs. The new smoke step is gated to skip on sqlite rather than paint that matrix leg red; postgres runs on every NC version, so `forUpdate()` stays covered.

## Also caught
`docs/RELEASING.md:154` carried a release-checklist assertion that `min-version="30"` — it would have failed the next release. Stale range claims in `README.md`, `CLAUDE.md`, `dev/docker-compose.yml` and `dev/setup.sh` updated too, and `CONTRIBUTING.md` gained a short note on *why* only the install matrix can catch this class: the `nextcloud/ocp` pin means psalm and PHPUnit always check against the newest API, never the oldest supported one.

Dropping the two EOL legs also removes 4 of 11 install jobs, so CI gets faster.

No app `<version>` bump, no `CHANGELOG.md` edit. `info.xml` validated against the App Store XSD.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -317,10 +317,11 @@ jobs:
         run: npm run l10n:lint
 
   # Cross-version × cross-DB install/migration matrix. This is what actually
-  # VALIDATES the info.xml claim of NC 30–34: for each (version, db) combo it
+  # VALIDATES the info.xml claim of NC 32–34: for each (version, db) combo it
   # builds the frontend, boots a throwaway Nextcloud on that DB, enables Kanso
   # (running every migration via the app-enable + `occ upgrade`), then asserts
-  # the schema was really created (dev/smoke.sh). It deliberately does NOT run
+  # the schema was really created AND that app code actually runs against that
+  # server's API (dev/smoke.sh). It deliberately does NOT run
   # Playwright — the full browser e2e is slow and stays on 34+postgres only (the
   # `e2e` job below). PHPUnit is covered separately by the `unit-php` job (across
   # PHP 8.2 and 8.3); it is NOT re-run inside these containers because the ARC
@@ -340,7 +341,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        nc-version: ['30', '31', '32', '33', '34']
+        nc-version: ['32', '33', '34']
         db: ['sqlite', 'postgres']
         include:
           - nc-version: '34'
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,7 +1,7 @@
 # Kanso
 
 Kanso is an open-source kanban app for Nextcloud (app id `kanso`) — a from-scratch,
-performance-first kanban board. AGPL, targets Nextcloud 30–34.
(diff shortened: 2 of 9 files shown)


aktasfatih:
:tada: This pull request is included in version 0.23.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.23.0).

PR #107, opened 2026-09-04
Fix two shipped fatals, cap unbounded input, and add resolvable discussion threads

aktasfatih:
A `/pm` session drained six cards off the Kanso board onto one branch. Two shipped fatals, one hardening fix, one CI reliability fix, one App Store listing fix, and one feature.

## What's here

| Commit | Card | |
|---|---|---|
| `69cdd60` | #10149 | `fix(rebalance)` — repair sort-key rebalance on SQLite and on Nextcloud 32.0.0–32.0.6 |
| `c7a234f` | #10144 | `fix(cards)` — cap description length and bound how many mentions one save resolves |
| `96ce580` | #10168 | `ci(e2e)` — pre-pull the dev stack images with a retry |
| `90bbb16` | #10164 | `fix(appstore)` — list Kanso under Office & text as well as Organization |
| `2f772c9` | #10166 | `fix(cards)` — `maxlength` on the four card-title inputs |
| `2d6bbc1` | #10165 | `feat(comments)` — mark a discussion thread resolved to collapse it |

No `<version>` bump anywhere (release-only); the one migration lands unbumped.

## The two shipped fatals

**#10149** started as "`occ kanso:rebalance` fatals on SQLite". Validation found a second fatal on the *same line*: `IQueryBuilder::forUpdate()` is `@since 33.0.0` on master but only `@since 32.0.7` on `stable32`, while the declared floor is `32` — so `CardMapper.php:381` was a fatal `\Error` on NC 32.0.0–32.0.6 **on every database**. `install-matrix` can't see it because it pulls `nextcloud:32`, the latest patch. Both halves are now guarded (`getDatabaseProvider()` + `method_exists`), and dropping either one alone leaves the other case red — they're independently load-bearing.

The user-visible harm was never the `occ` exit code: on SQLite the *only* documented recovery from the 64-char sort-key wall was unavailable, so an overflowed stack was permanently stuck for that install.

Also here: the sqlite gate added to `dev/smoke.sh` by PR #106 is removed (it was hiding exactly what it should catch), `updateSortKeyById()` gained a `stack_id` predicate, and the now-false `min-version` rationale comment in `info.xml` is corrected — `forUpdate()` is `@since 32.0.7`, `PARAM_DATETIME_MUTABLE` is `@since 31.0.0`. **`min-version="32"` itself is unchanged.**

**#10144** capped descriptions at 65,536 characters. Two hypotheses that would have made the card moot were tested and both are false: the column is a length-less `Types::TEXT` (→ LONGTEXT/4 GiB on MySQL), and there was no frontend cap either. The card's own prescription needed three corrections — `CardService::create()` takes no `$description` at all, the importer paths must stay uncapped or Deck/Trello import of long descriptions breaks, and the mention bound belongs in `extractUsernames()` because `handleMentions` was *already* an unbounded ACL-query amplifier on the **comment** path (10,000 chars holds ~3,300 distinct `@tokens`).

## Verification

Every card was proved by mutation — break the guard, re-run, confirm red, restore. Highlights:

- **#10149** ran against four throwaway containers (NC 34 + sqlite, NC 32.0.6 + postgres, NC 34 + postgres, NC 34 + mariadb), with both fatals reproduced as red baselines first. Postgres `log_statement='all'` confirms `… ORDER BY "sort_key" ASC FOR UPDATE` is *still issued* after the fix — the mysql/postgres posture is not weakened.
- **#10144**: deleting the mention cap makes a test count **5,000 real ACL lookups instead of 50** — it invokes `getPermissions` rather than stubbing it.
- **#10168**: the step's `run:` block was extracted from the parsed YAML and executed verbatim under the **CI-pinned** Compose v2.29.7. This caught a trap: `docker-compose.yml` declares `env_file: .db.env`, which `setup.sh` only writes when it runs, so a naive pre-pull dies with exit 14 *before pulling a byte* — while host Compose v5.1.1 tolerates it, meaning the broken version would have looked fine locally and been dead on the runner.
- **#10164**: valid category slugs confirmed against two independent sources that agree exactly — the live App Store API and the XSD vendored at `scripts/schema/info.xsd`. The repo's own `lint:info-xml` guard was run *and* proved non-vacuous (`officex` → enumeration error).
- **#10165**: 3 new e2e green, 19 in a regression sweep, 7 more on activity/mentions; seven mutation proofs including "never collapse" and "no deep-link expand", each verified red through a rebuilt frontend.

## Known limits, stated plainly

- PHPUnit mocks `IDBConnection`, so #10149's mapper tests pin *which calls the mapper makes per provider*, not what the database does. That SQLite really rejects `FOR UPDATE`, and the `method_exists` half, are covered only by the container runs and the install matrix. This is written into the test file.
- #10144's cap bounds what a user **types**, not every row — copy/template/import paths still carry long descriptions across by design, so the search amplifier is not fully bounded. Tracked as a separate board card with a steer toward fixing it at the search layer (a SQL-side `SUBSTRING`) instead of policing every writer.
- Workers ran targeted e2e specs (~30 tests), not the full ~1.5h suite. CI owns that.
- #10165's export/import round-trip is PHPUnit-only. The local PHP runner has no `ext-zip`, so the archive/backup suites are red on clean `HEAD` too and could not run.

## Reviewer notes

- **#10165 includes one deliberate addition beyond its spec** (~2 lines + 2 assertions): `ExportService` emits `resolvedAt` and `ImportService` reads it back. Without it, a board export/import round-trip silently reopens every settled thread — a data-loss defect the feature itself would have introduced. Both directions are mutation-proved; an older archive without the field imports as open.
- In `CardDetail.vue`, the expanded-thread branch is wrapped in `<template v-else>` **without re-indenting its 165 lines**, to keep the diff readable. Functionally correct; the re-indent is worth a separate whitespace commit if you want it.
- `dev/setup.sh` is deliberately untouched throughout — there is uncommitted local work in it — which is why the pre-pull retry lives in `ci.yml`. That constraint is recorded in the commit message and in an in-file comment so it doesn't get "tidied" into the script later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -407,6 +407,76 @@ jobs:
           fi
           docker compose version
 
+      # Warm the local image cache BEFORE booting the stack, retrying a failed
+      # pull, because `dev/setup.sh` cannot: it boots with `docker compose
+      # ... up -d` under `set -eu` (dev/setup.sh:19,67), where the pull is
+      # implicit and has NO retry. One transient registry hiccup therefore kills
+      # the whole job — observed on PR #101, run 33560679687, 6m46s into `e2e`
+      # (the ~1.5h required check) after five minutes of layers downloading
+      # fine: `Get "https://registry-1.docker.io/v2/": net/http: request
+      # canceled while waiting for connection (Client.Timeout exceeded while
+      # awaiting headers)`, exit 18. That red is indistinguishable from a real
+      # regression until a human opens the log and re-runs by hand.
+      #
+      # Deliberately HERE and not inside dev/setup.sh: the script is the shared
+      # local-dev entry point and stays free of CI-only retry plumbing. Please
+      # do not "tidy" this into it.
+      #
+      # Also deliberately NOT a retry around `./dev/setup.sh` itself — re-running
+      # that script after a partial boot is not a claim this makes. Only the
+      # network fetch is retried, and a fetch is safe to repeat.
+      #
+      # Best-effort by construction (`exit 0` at the end, plus
+      # continue-on-error): this warms a cache, it does not gate anything. If
+      # all three attempts fail, `up -d` still gets its own try and any honest
+      # failure surfaces there, under the step whose name says "Boot".
+      - name: Pre-pull dev stack images (with retry)
+        working-directory: dev
+        # Belt and braces with the `exit 0` below: this step must not be able to
+        # redden the job even if the script itself breaks before reaching it.
+        continue-on-error: true
+        run: |
+          set -u
+          # Load-bearing, not defensive. docker-compose.yml declares
+          # `env_file: .db.env` on the nextcloud service, and setup.sh only
+          # writes that file when IT runs. Compose v2.29.7 — the version the
+          # step above pins for runners without the plugin — refuses to load
+          # the model without it and dies
+          # with `env file .../.db.env not found`, exit 14, BEFORE pulling a
+          # single byte. So without this placeholder all three attempts below
+          # would fail instantly and this whole step would be silently vacuous.
+          # (Compose v5 tolerates a missing env_file for `pull`, so this would
+          # look fine on a newer runner and do nothing on the pinned one — the
+          # worst kind of CI bug.) setup.sh overwrites the file moments later
+          # with the real driver env.
+          [ -f .db.env ] || : > .db.env
+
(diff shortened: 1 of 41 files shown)


aktasfatih:
:tada: This pull request is included in version 0.24.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.24.0).

PR #110, opened 2026-09-04
Fix five shipped defects, and add the CI job that would have caught a class of them

### LINKED ISSUE #108: [Bug] Date in Activity tab is displayed twice (only on existing/older cards)
Hi @aktasfatih,

The one of recent update to show exact timestamps in the activity log works great for new cards, but it introduced a small formatting bug on existing ones.

On cards created before the update, the date is now displayed twice next to each activity (e.g., "26. Aug. 2026 · 26. Aug. 2026, 06:21"). Newly created cards are not affected and display correctly.

It would be great if you could check this.👍

aktasfatih:
A `/pm` session off `v0.24.0`. Five shipped defects users can feel, plus the CI job that would have caught a whole class of them.

## What's here

| Commit | Card | |
|---|---|---|
| `e712672` | #10177 | `fix(activity)` — stop repeating the date on entries older than a week |
| `70a5456` | #10146 | `fix(pwa)` — stop offline board data reaching a different user on a shared device |
| `475d366` | #10171 | `fix(card)` — keep the editor open with your text when an edit fails |
| `ead8cb0` | #10127 | `fix(reminders)` — no due reminders for cards on archived or trashed boards |
| `6288f3f` | #10174 | `ci` — test the upgrade path: install the last release, seed it, then upgrade |
| `48ea27f` | #10180 | `fix(reminders)` — no due reminders for cards saved as templates |

No `<version>` bump anywhere (release-only). No migrations.

## The one to read first: `6288f3f`

Every CI job installs the branch's own tree fresh, so **every migration has always met an empty database**. Nothing installs a released Kanso, seeds it, and upgrades.

The new `upgrade` job does that, and it was proved both directions:

- **Non-vacuity** — a probe migration adding a `NOT NULL` column with no default to `kanso_comments` turns the job **red**: `SQLSTATE[23502]: Not null violation: column "probe_notnull" … contains null values`.
- **Complement** — the *same commit, same migration*, installed the way every existing job does: `setup.sh` exit 0, `smoke.sh` → `SMOKE OK`, column present. **The current matrix would have shipped it green.**

Two traps it had to solve, either of which would have made the job worthless:
1. Versions are release-only, so the branch carries the *same* version as the last release and `occ upgrade` would have been a **no-op** — permanently green and meaningless. The script stamps the *staged copy* (never the checkout) with the next patch version, mirroring semantic-release, then asserts `installed_version` actually moved.
2. Right after `occ app:enable`, apache workers still 404 the app's routes (CLI vs. worker app-config cache). It failed a run at `POST /api/boards -> HTTP 404`; fixed with a bounded `wait_for_api`. The same race latently affects `dev/smoke.sh` in `install-matrix` — **not** touched here.

Runtime **44–46 s** locally, `timeout-minutes: 30`. `dev/seed.sh`'s header lists exactly which tables it seeds **and which it does not**; CONTRIBUTING.md tells migration authors to extend it.

⚠️ The `upgrade` job is **not** in the branch-protection required list — that's a repo setting a maintainer must flip. Noted in the workflow and in CONTRIBUTING.md.

## The privacy fix: `70a5456`

Filed as an offline-only service-worker leak. It is worse than that: `src/services/offlineCache.js` hydrates IndexedDB **before mount** with no identity check, so on a shared browser profile user B's first paint rendered user A's boards **while online**.

Fixed both halves — `/api/` GETs now bypass the worker cache, and the persisted snapshot is stamped with the owner's uid and deleted rather than hydrated on mismatch. The worker `VERSION` bump v3 → v4 is load-bearing, not cosmetic: `activate` purges caches without the current prefix, so upgrading also clears API payloads a v3 worker already wrote on existing installs.

**Honest limit, documented in the code and tracked separately:** the *offline* path is not closed. Offline, the cached app shell still carries the previous user's `data-user`, so the identity check agrees with the stale snapshot. No offline client can verify a session it cannot reach. Closing it needs a design decision (identity-free offline shell, session-derived encryption, or document-and-accept) — it is not an oversight.

## Verification

Every card was proved by mutation — break it, re-run, confirm red, restore. The two reminder fixes could not be proved by PHPUnit at all (`DueReminderServiceTest` mocks `CardMapper`; there is no DB-backed test layer), so they were proved end-to-end against seeded data on a real stack:

- **#10127** — archived board → **0**, trashed board → **0**, active control → **1**, with SQL confirming all three *cards* were still `archived='f', deleted_at=0`. Removing the predicate: `1 / 1 / 1`.
- **#10180** — template → **0**, control → **1**, and the template row stayed unstamped (`due_reminder_sent = 0`), confirming SQL exclusion rather than a PHP-side skip. Removing it: `1 / 1`.

The unit tests for both pin only that the predicate is **emitted**, and their docblocks say so.

- **#10171** — restoring the three `finally` blocks: **3 failed, 1 passed** — every failure case red, the success-path control still green.
- **#10146** — disabling the SW guard refilled the cache with `/api/boards/29`, `/participants`, `/saved-filters`; disabling the identity check rendered another user's card (`Expected: 0  Received: 1`).
- **#10177** — reproduced the reporter's exact string on revert: `"27. Aug. 2026 · 27. Aug. 2026, 13:26"  Expected: 1  Received: 2`.

## Reviewer notes

- **#10127** carries two deliberate calls: `archived = false` is left unguarded against NULL (repo idiom, no backfill window, 0 of 52 live rows NULL), and un-archiving a board **replays** its accumulated reminder backlog — chosen over stamping skipped cards, which would silently consume a reminder the user never saw. Both are documented in the docblock.
- **#10177** has no permanent e2e for the ≥7-day case: aging an activity entry needs a direct `UPDATE` on `kanso_changes`, which would not survive the sqlite/mysql legs. The browser proof was done and thrown away; the unit test is the durable gate.
- **#10180** corrects a false citation inherited from its own card — `CardService.php:685` does *not* copy `duedate` into templates (that is `buildMovedCard()`, which sets `isTemplate(false)`). The real route in is `setTemplate()`, a pure flag flip on an already-dated card.

Closes #108


aktasfatih:
:tada: This pull request is included in version 0.24.1 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.24.1).

PR #111, opened 2026-09-04
feat: Forgejo integration (receive-only), plus a GitHub PR-state fix

aktasfatih:
Adds a Forgejo (and Gitea) integration alongside the existing GitHub one, in response to a user request.

## Design: receive-only

The obvious shape — mirror the GitHub feature, add a per-board instance URL — was **dropped after looking at what the URL is actually for**. Kanso never dials out for a webhook; it only ever receives. Auto-move keys off the `kanso-<id>` head branch and off matching the delivery's `html_url` against attached links. Neither needs to know where the instance lives.

The only thing that ever wanted a host is the read-time poll — and on self-hosted Forgejo that poll is close to worthless: most instances are private, so an unauthenticated request 404s and the badge stays `unknown` regardless. Making it work needs a stored **token**, not a URL.

So: **no instance URL, no token, no outbound HTTP.** Badges come from the deliveries, which are authoritative and free.

The trade-off is explicit and stated in the UI: *a link pasted for something that never changes again stays without a status*, because no delivery will mention it. If someone asks, the fix is a per-board token — purely additive, no migration rework.

This also removes the SSRF surface, the admin-allowlist question, and all the URL-normalisation edge cases (subpaths, ports, trailing slashes) that a stored base URL would have brought.

## A GitHub bug found on the way

`handlePullRequestEvent` called `addLink()`, which polls **only on INSERT**. The second delivery for a PR (opened → merged) hit the unique-constraint branch and refreshed nothing — so **a PR's "merged" badge was never set by the merge webhook**, only by whenever someone next opened the card and the 300s poll happened to succeed. The delivery carrying `merged: true` was discarded.

That is also why `card-links.spec.js:51` could only assert `STATES.toContain(state)` — the state genuinely wasn't deterministic.

Fixed first, as its own commit, on its own merit.

## Commits

1. `fix:` cache PR state from the payload; read a merge from all three spellings (`merged` bool / `merged_at` / `merged` action) since a merged PR reports `state: closed`
2. `refactor:` extract the provider-agnostic half into `AbstractForgeWebhookService` driven by a normalized `ForgeEvent`
3. `feat:` a `reason` enum on no-op deliveries (`no_branch_match`, `intake_filtered`, …) — a bare `handled: false` is undiagnosable from the delivery log
4. `feat:` the Forgejo service, endpoints, migration and UI

## Notes for review

- **The refactor is the risk**, so it was gated: mechanical extraction only, and `GithubWebhookServiceTest.php` was **not opened** during it — 47/47 passed byte-identical before and after. A base class rather than an injected collaborator specifically so the constructor signature (and thus the test wiring) stayed unchanged.
- **Forgejo signatures** are a raw hex digest, not `sha256=<hex>`. A hook saved *without* a secret still sends the header **empty**, so signatures are rejected on shape (exactly 64 lowercase hex chars) before any comparison — empty falls out as malformed rather than as "no signature offered".
- **Issues and PRs share one number sequence** and redirect to each other, so both spellings are match candidates: a link pasted as `/issues/5` is still updated by a `/pulls/5` delivery. Without this the badge would sit on `unknown` forever.
- **A foreign-host link is accepted only once the board's Forgejo webhook is enabled.** Kanso stores no instance URL, so that opt-in is the only available signal; boards that haven't opted in keep the strict github.com-only rejection.
- **Version drift fails safe**: an unrecognized action yields no target role, so the delivery is an accepted no-op. Drift costs a *missing* move, never a wrong one. All tolerance lives in one normalizer.
- `provider` on `kanso_card_links` backfills to `github` via the column default; the poll is now gated on it, so a self-hosted instance is never contacted.
- No version bump (release-only, per CLAUDE.md).

## Verification

Run against the repo in a throwaway container (the dev stack is serving a different worktree):

- `phpunit` — **1446 tests green** (was 1404; +42)
- `psalm` — no errors
- `php-cs-fixer --dry-run` — 0 of 371 files need fixing
- `vite build` — clean
- `l10n:extract` + `l10n:compile` re-run as the last step

The `fix:` commit was checked for **non-vacuity**: disabling it makes the new test fail with `update() expected 1 time, actually called 0 times`.

⚠️ **`tests/e2e/card-forgejo-webhook.spec.js` is authored but was NOT executed locally.** The dev stack mounts a different worktree and `compose down` resets its data, so running it would have meant tearing down work in progress. CI's `e2e` job covers it — but it is unrun, so treat a failure there as expected-to-be-real rather than flake.

---

## Rebased onto main (after #110)

Main moved 181 commits while this was open. #110 had substantially rewritten `handlePullRequestEvent`, so the rebase was **not** mechanical — replaying the refactor as-authored would have silently deleted three behaviours main added. Each was re-extracted into the base class instead:

- **`PREFIX-<seq>` title references** (#9855) — branch **and** title matches unioned into a `targets` map with per-card echo flags, plus the refined egress rule where a title-derived id is *not* an echo (it would leak the seq → id mapping). **Forgejo inherits this for free**: the reference lives in the PR title, which every provider's payload carries.
- **Form-encoded body rejection** — a correctly-signed non-JSON body throws `NonJsonWebhookBodyException` → 400 with the content-type hint, rather than a green 200 that does nothing.
- **Stack lookup inside the try** in `applyAutoMove` — a title can match several cards, so a DB error there would escape as a 5xx and disable the hook.

The `GithubWebhookServiceTest` suite caught all three (it grew from 47 to 69 tests on main); nothing else in it needed editing.

Other rebase resolutions:

- **Migration renumbered `005600` → `005900`** — main claimed 005600–005800.
- **`REASON_NO_BRANCH_MATCH` → `REASON_NO_CARD_MATCH`**, since targets now come from the branch *or* the title.
- **`CardDetail.vue`** keeps main's `v-if="cardFeatures.github"` per-board feature switch (#5894). The stored key stays `github` for existing boards, and now gates the whole section — both forges. Renaming the key would need a data migration for no user-visible gain; flagging in case you'd rather it read `code`.
- **README** keeps main's `application/json` content-type note.
- `.pot` regenerated rather than hand-merged.

**Re-verified on the rebased HEAD:** `phpunit` **1823 green**, `psalm` 0 errors, `php-cs-fixer` 0 of 389, `vite build` clean.

The e2e spec caveat above still stands — still authored but not locally executed.


aktasfatih:
:tada: This pull request is included in version 0.26.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.26.0).

PR #112, opened 2026-09-05
Eight board cards: two MCP surfaces, a real checklist bug, and four tests that could not fail

aktasfatih:
The theme is **things that were quietly not true** — two tests that could not fail, an assertion pointing at the wrong field, a guard that rejected valid input, and a UI element that had been misaligned since it shipped. Every fix here was proven by mutation (break it, watch the test go red, restore) rather than by "it passes now".

### Two MCP capabilities you asked for
- **`feat(mcp)` — see your saved cross-board Views from an AI assistant.** Read-only by explicit decision: `kanso_list_views` + `kanso_get_view_cards`. No create/update/delete, and a test now pins that — a future `kanso_save_view` turns it red.
- **`feat(mcp)` — move a card to another board.** The server side already shipped; only the tool was missing. The docstring spells out the lossy semantics (new card id, comments/attachments left behind) because that is the only thing stopping a model from silently stranding a user's data.

### A real bug found by de-vacuuming a test
- **`fix(checklist)` — stop dropping the tick on an item you just added.** Chasing a 4×-retry loop in the e2e suite turned up an actual product defect: `addItem` inserted an optimistic row with a negative placeholder id, and a click before the settle refetch sent `PATCH /api/checklist/-1788…`, which rolled back silently. Adding a checklist item and immediately ticking it lost the tick. Fixed in `useChecklist.js` + `CardDetail.vue`; the retry loop that had been hiding it is gone.

### Tests that could not fail
- **`test(e2e)` — make the public-share leak guard real.** The spec posted `{message: …}` where the API takes `body`, so no comment was ever created and the "does a public share leak internal comments?" assertion was vacuous. Now proven: with a deliberate leak the payload comes back containing the secret and the test goes red. Also attributed the long-standing local-only failure to three causes (missing frontend build, token minted inside a test, and Nextcloud brute-force throttling) and fixed all three; `dev/setup.sh` now builds and resets the throttle.
- **`test(e2e)` — let the inbox, checklist and timeline specs actually fail.** `inbox.spec.js` swallowed its ACL setup and degraded four tests to skips, so a real permission regression passed green. `timeline-view.spec.js` had an optional marker assertion whose else-branch asserted nothing new. Both now fail when broken — demonstrated by mutating each.

### Correctness and hardening
- **`fix(recurrence')` — reject a repeat whose months are too short for the day it starts on.** A `FREQ=YEARLY;BYMONTH=2` rule anchored on the 31st silently fired on 3 March and then drifted permanently. The overflow is inside `Sabre\VObject`, which resolves from the *host* Nextcloud's tree and is not ours to patch, so this rejects at validation with a 400 naming the months that don't fit. Checks each listed month individually — `BYMONTH=1` on a 31-Jan anchor is still accepted.
- **`fix(security)` — cap a card's GitHub links and resolve Deck attachments by name only.** A per-card link cap (unbounded links meant a reader could tie up a PHP worker for N×4s), and Deck's raw filename now goes through `basename()` before the storage lookup.
- **`fix(editor)` — remove the stray attributes the description editor put on its root element.** The root `<div>`'s opening tag was never closed, so a comment got swallowed into it as four garbage attributes. Verified cosmetic, not a throw.
- **`fix(ui)` — line the My Work pages up with the hub header.** Each embedded page re-applied the hub's own padding *and* max-width, leaving content 32px out of line with the header above it — and narrow enough to ellipsise card titles.

### Notes for review
- No `<version>` bump anywhere (release-only, per `CLAUDE.md`).
- Gates run locally: PHPUnit **1789 passed**, MCP pytest **78 passed**, psalm clean, php-cs-fixer clean, `l10n:check` clean, and the touched e2e specs green at `--retries=0` under `E2E_ISOLATE=1`.
- `dev/setup.sh` gains a build step, so CI now builds twice (~11s/job). Set `KANSO_SKIP_BUILD=1` on the e2e job if that's not worth it — `ci.yml` was deliberately left untouched.
- Three follow-ups were filed on the board rather than smuggled in here: the public-share guard's substring-vs-identity check, `GithubWebhookService` bypassing the new link cap, and an unscoped `.empty-content` selector that matches Nextcloud's own chrome.

---

## Cycles 2–3 (added after the PR opened)

- **`fix(ui)` — long names no longer silently cut or rejected.** 17 uncapped name inputs got a `maxlength` matching *their own* entity's cap (13×100, 4×255), caps verified against both the service constant and the migration column width. The view-name and saved-filter fields came first: those were the only two paths that `mb_substr`-truncated with no error, i.e. silent data loss.
- **`perf(ics)` — unchanged calendar-feed polls answer 304 instead of rebuilding the feed.** The card asked for an `AnonRateLimit`; that was declined (IP-keyed, and one NAT'd office of calendar clients would trip any safe threshold). Reusing the app's own ETag bet makes the anonymous endpoint cheap rather than gated, and the query is now bounded. Verified with a live curl probe: 200+ETag → 304 empty → 200 on stale → 304 on the weak form → 404 on an unknown token.
- **`fix(security)` — per-user rate limits on CSV import and both attachment upload paths.**
- **`perf(search)` — search no longer hydrates whole descriptions.** `searchInBoards` selected `*`, pulling up to 100 descriptions of up to 32 MiB each to build 160-char snippets; it now clips in SQL. No writer was capped — that bypass is deliberate and documented, and a mutation test now catches anyone who "fixes" it that way.
- **`fix(ui)` — a filter on a facet the board no longer offers can be cleared again.** The card called this URL-only; it is not. Delete a board's last label while filtered by it and the board renders zero cards with a filter badge and no clear button. Fixed without reverting #10091's invariant.
- **`test(filters)` / `test(e2e)`** — pinned the two filter functions nothing tested (a miss there returns *unfiltered rows*), and scoped the Views spec so a large database can't hide its own fixtures.

### One thing to know while reviewing
`tests/e2e/waiting-on-client.spec.js:113` fails **locally** — on `origin/main` too. It is not this branch and not a product bug: the local dev stack has `notify_push` enabled with an unreachable daemon, which stretches the delta poll past the spec's timeout. CI never installs that app. Bisected and confirmed; filed as #10225 (the real product defect it exposed — a dead push daemon silently degrades every board to a 30s refresh) and #10226 (dev-stack fix).


aktasfatih:
:tada: This pull request is included in version 0.25.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.25.0).

PR #113, opened 2026-09-08
fix: match both issue and pull-request URL spellings on forge links

aktasfatih:
Follow-up to #111, closing the gaps I flagged when that merged.

## What changed

**`fix:` — a PR link pasted with the issue spelling now stays fresh.**

A forge shares one number sequence between issues and pull requests and redirects each spelling to the other, so pasting `…/issues/5` for a pull request is a legitimate thing to do. The delivery then carries `…/pull/5`, which matched nothing — so `addLink` attached a **second** link beside the pasted one, and the pasted one sat on `unknown` permanently.

Both spellings are now candidates. An existing link is refreshed in place rather than duplicated, and an `issues` delivery matches a link stored either way. This was GitHub-only; Forgejo already matched both spellings, which is what made the GitHub side visible by comparison.

**`test:` — the shared title matching is now pinned for Forgejo.**

`PREFIX-<seq>` matching lives in `AbstractForgeWebhookService`, so Forgejo has had it since #111 landed — but only the GitHub suite covered it, leaving the Forgejo half free to regress silently. Adds the branch/title union, the case-sensitivity rule that stops a quoted `kanso-<id>` branch being read as a seq reference, and the unresolvable-reference no-op.

## One thing I did NOT do

I had also flagged that **issue intake bypasses the per-card link cap**. On re-reading the merged code that claim is wrong: `intakeIssue` creates a brand-new card and puts exactly one link on it, and that is the only direct mapper insert in the base class. A fresh card has zero links, so `MAX_PER_CARD` is unreachable there. Adding a guard would have been dead code, so there is no change for it.

## Verification

- `phpunit` **1859 green** (+7)
- `psalm` 0 errors, `php-cs-fixer` 0 of 394
- No UI strings touched, so no l10n or bundle changes

Both behaviours were **proved by mutation**, not just asserted:

- forcing the alternate-spelling lookup to find nothing → `testPrDeliveryRefreshesALinkPastedWithTheIssueSpelling` fails
- narrowing the candidate set back to `issues` only → `testIssueDeliveryMatchesALinkStoredWithThePullSpelling` fails

### DIFF
--- a/lib/Service/AbstractForgeWebhookService.php
+++ b/lib/Service/AbstractForgeWebhookService.php
@@ -324,14 +324,7 @@ abstract class AbstractForgeWebhookService {
 			// poll: addLink() only polls on INSERT, so the second delivery for a
 			// PR (opened -> merged) hits its unique-constraint branch and would
 			// otherwise never refresh the state.
-			if ($event->htmlUrl !== '') {
-				try {
-					$link = $this->cardLinkService->addLink($cardId, $event->htmlUrl, $board->getOwner());
-					$this->cacheLinkState($link, $event);
-				} catch (\Throwable) {
-					// Non-critical - a bad/duplicate URL must not fail the webhook.
-				}
-			}
+			$this->attachOrRefreshPrLink($cardId, $event, $board->getOwner());
 			[$cardMoved, $reason] = $this->applyPrAutoMove($boardId, $cardId, $event, $board->getOwner());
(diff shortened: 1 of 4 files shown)


aktasfatih:
:tada: This pull request is included in version 0.26.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.26.0).

PR #116, opened 2026-09-09
fix: realtime polling honesty, sub-day recurrence offsets, and a stricter public-share guard

aktasfatih:
Autonomous /pm session on the Kanso board (board 19). One branch, one PR, **18 cards, 22 commits**.

> ### ⚠️ Contains three security fixes — please prioritise review
> - **`36523c7`** — anonymous share links were leaking real Nextcloud **login uids** through `@mentions` in card descriptions, in the **default** config (comments off). Reproduced live before the fix.
> - **`b46136f`** — CSV import minted board-level labels at **EDIT**, bypassing `LabelService`'s **MANAGE** gate; EDIT also survives the external-role strip.
> - **`87264bd`** — `ReviewService::setState` let any READ member record a verdict on **someone else's** review, and let a non-member distinguish 403/404 to probe review existence.
>
> If you would rather these went out as their own small PR, say so on the board and I will split them.

> ### Also: PR #109 is superseded — please CLOSE it rather than merge
> `5e86d08` is its commit, byte-identical. #109 had been open since 2026-09-04 with every required check green, blocked only by a `.po`/`.pot` line-ref conflict. Merging both would land the same fix twice.

## Theme, which emerged rather than being planned

**Guards and assertions that could not fail.** It started with one flaky test and ended up recursive — the session found this defect class in shipped code, in the tests written to prove the fixes, and in its own tooling.

## What landed

| commit | card | what |
|---|---|---|
| `5e86d08` | 10276 | `fix(recurrence)` sub-day due-date lead time survives editing a repeat rule |
| `672336e` | 10225 | `fix(realtime)` boards stay on the fast 5s refresh until push proves it works |
| `8eba10b` | 10222 | `test(e2e)` pin the exact public-share payload key sets |
| `b63826a` | 10278 | `fix(realtime)` stop polling a board while its tab is backgrounded |
| `5f490d8` | 10226 | `chore(dev)` verify notify_push on boot instead of assuming it |
| `e54af07` | 10255 | `feat(boards)` a board can hide the checklist section like the others |
| `e779f3a` | 10284 | `fix(checklist)` a just-added step no longer loses its assignee on a slow connection |
| `327e2ba` | 10285 | `fix(checklist)` a just-added step no longer reappears when you delete it |
| `a50b67c` | 10275 | `fix(ui)` hover hint on every button in the multi-select action bar |
| `59feaf3` | 10281 | `chore(dev)` pin both halves of notify_push so the boot check can't cry wolf |
| `59b7309` `98095dd` | 10292 | `test` + `fix(realtime)` make three shipped guards fail when removed; no orphaned poll loop |
| `36523c7` | 10294 | **`fix(share)` display names, not login names, in public board text** |
| `b46136f` | 10295 | **`fix` CSV import label creation now needs MANAGE** |
| `87264bd` `e13d400` | 10296 | **`fix(security)` + `test` review verdicts are the reviewer's alone; no existence oracle** |
| `9871ce0` | 10297 | `fix(import)` show what an import copied, and warn about anything it skipped |
| `c4df0ca` | 10305 | `test(e2e)` probe the review routes in the visibility leak matrix |
| `d7bb6c0` `36b77ad` | 10310 | `fix(comments)` + `test(e2e)` a deep-linked comment stays in view while the card loads |

## Review notes

- **No `<version>` bump** anywhere — versioning is release-only here.
- **Please merge with a merge commit.** Commit types are deliberate (`fix:` ×11, `feat:` ×1, `test:` ×3, `chore(dev):` ×2); a squash under one title would collapse them and lose release entries.
- **Every behavioural change is mutation-proven** — break the fix, confirm red, restore. `327e2ba` carries nine such proofs, `87264bd` nine, one per guard.
- Three of my own specs were **wrong and the workers said so** rather than building them: an `apache2ctl` root cause that did not exist (and whose "fix" broke the boot), a mutation proof that was impossible once the fix landed, and a leak-matrix probe that **could not have failed** because its viewer was a board member. Those corrections are recorded on the cards.
- **The earlier e2e failure was attributed and is not from this branch**: branch and `main` bundles fail and pass at identical latency thresholds across six injection points. Investigating it is what surfaced the two `fix(checklist)` bugs.
- **23 follow-up cards** were filed rather than scope-crept into these commits, including two that correct /pm's own earlier reasoning.

## On the one CI failure, and what it found

Run `34358052981` on `c4df0ca` was **14/15 green** — full install matrix (NC 32/33/34 × sqlite/postgres/mysql), upgrade-with-data, psalm, cs-check, both PHP versions, build-frontend, unit-mcp — with `e2e` failing 2 of 599.

Both failures were attributed by bundle-swap A/B (identical backend and fixtures, only the built JS swapped) as **pre-existing, not from this branch**:

- `accessibility.spec.js` — a test race. `src/views/BoardView.vue` is **not in this branch's diff at all**; `.board-view` renders skeleton columns before the board read resolves, and a 4s API delay reproduces the CI output character-for-character. Fixed by waiting for content before snapshotting; **the assertion is unchanged**.
- `comment-deeplink.spec.js` — **a real product bug, now fixed.** Under CPU throttling it fails **0/3 on `origin/main`** and 3/3 passes only after the fix. Measured cause: the thread pane *shrinks after* the scroll (the composer and its lazy `MarkdownEditor` mount once the board read resolves), `scrollTop` survives but the centred comment does not, and `scrollHandled` is already latched. **User impact: a reminder or mention link silently lands the reader on a neighbouring comment.** Fixed with a `ResizeObserver` that re-centres on resize only, so hand-scrolling isn't fought.

Neither assertion was weakened. Both fixes are mutation-proven, and the specs were run 3× at workers=1 and 2× at workers=2 (CI's setting).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -694,8 +694,9 @@ jobs:
       # postgres + redis + notify_push + nextcloud:34-apache, app enabled,
       # firstrunwizard disabled, tester user + push wired up. Idempotent.
       #
-      # setup.sh also side-loads the optional apps two specs need — deck (for
-      # deck-import.spec.js) and contacts (for card-contacts.spec.js) — via
+      # setup.sh also side-loads the optional apps a few specs need — deck (for
+      # deck-import.spec.js and deck-import-ui.spec.js) and contacts (for
+      # card-contacts.spec.js) — via
       # dev/install-optional-apps.sh. That used to be two steps here, which
       # meant a local stack silently lacked both and those specs were red for
       # everyone but CI; the pins now live in the script, in one place. It still
--- a/README.md
+++ b/README.md
@@ -230,7 +230,20 @@ is **not** bundled into the installed Nextcloud app.
 
 The dev stack mounts your checkout as `custom_apps/kanso`, so a rebuild
 (`npm run build`) plus a browser reload picks up frontend changes; PHP changes
-apply immediately. `docker compose down` in `dev/` resets the instance.
+apply immediately. In `dev/`:
+
+```sh
+docker compose down && ./setup.sh                    # restart: install + data survive
+docker compose --profile '*' down -v && ./setup.sh   # reset: wipes everything
+```
+
+`--profile '*'` on the reset matters — `-v` only removes the volumes of services
+in active profiles, so a bare `docker compose down -v` deletes the webroot but
+leaves the database, and the next boot fails trying to install over it. Reset
+(not restart) after switching `NC_VERSION` or `KANSO_DB`, and after **adding a
+migration**: `setup.sh` only re-enables the app, and Nextcloud re-runs an app's
+migrations only when `info.xml`'s `<version>` grows, which feature branches never
+do. `./seed.sh` refills a reset stack; `./smoke.sh` reports a schema left behind.
 
 PHP tooling runs via Docker (no host PHP needed):
 
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -81,7 +81,25 @@ services:
   # for postgres (its DATABASE_URL is postgres-specific and realtime push is
   # not exercised by the cross-version install matrix).
   notify_push:
-    image: icewind1991/notify_push
+    # Pinned deliberately, and it needs a MANUAL bump — in lockstep with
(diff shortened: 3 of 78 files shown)


aktasfatih:
:tada: This pull request is included in version 0.27.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.27.0).

PR #118, opened 2026-09-10
feat(mail): create cards by sending email to a board

### LINKED ISSUE #117: [feature request] Adding cards from e-mail
Hi,

It would be great if I could create new cards in Kanso via email. Meaning I send a message to a dedicated e-mail address, and a new card is then created on a specific board (e.g., the INBOX).

Best regards,
Jacek

aktasfatih:
Closes #117.

A board can point at an IMAP mailbox; every message that arrives becomes a card,
subject as the title and body as the description. Configured per board under
**Board settings → Automation → Email intake**, manage-only, polled every five
minutes.

## Why it looks the way it does

Nextcloud receives no mail, and the runtime ships neither `ext-imap` nor any IMAP
client or MIME parser in `3rdparty` — and Kanso ships no `vendor/`. So both are
hand-rolled on builtins: `ImapClient` speaks the slice of IMAP4rev1 this needs
over an `ImapTransport` interface (so the protocol is testable without a
network), and `MimeParser` decodes headers and walks the part tree using
`iconv_mime_decode`, `quoted_printable_decode` and `mb_convert_encoding`.

`EXAMINE` + `BODY.PEEK[]` throughout, so polling never writes flags on a mailbox
a person also reads. Intake position is a **UID watermark paired with
UIDVALIDITY**, not the `\Seen` flag — someone reading the mailbox in a mail
client cannot make the poller skip or repeat messages.

## This is the first path where a stranger causes a write

Every other way a card is created needs a session or an HMAC secret. Email intake
does not, and the text lands in a card body colleagues read. Most of this change
is that fact rather than the IMAP plumbing. The full threat model is in
`docs/email-intake-security.md`; the load-bearing ones:

- **`@mention` notification spoofing.** A description write re-parses `@name` and
  sends real notifications attributed to the writer. Intake writes a stranger's
  text *as the board owner*, so `@alice` in an email would have pinged Alice with
  the owner's name on it. Descriptions now write with `notifyMentions: false`.
- **SSRF.** `host`/`port` are free text from any board *manager*. `MailHostGuard`
  refuses loopback, private, link-local, CGNAT, multicast and IPv4-mapped
  equivalents; it resolves the name and the socket dials *that address*, closing
  the DNS-rebinding window, while the hostname travels separately so TLS still
  validates.
- **Auto-responder loops.** A card raises a notification email; an out-of-office
  answers it; that answer would have become the next card, forever.
  `Auto-Submitted`, `Precedence`, `List-*`, null `Return-Path` and delivery
  reports are never carded.
- **Credential theft by repointing.** A blank password field means "keep the
  stored one" — which let a manager who never knew the credential aim an existing
  mailbox at a host they control and collect it. A stored password is now only
  reusable for the same host and account.
- **Flooding.** Per-mailbox and per-sender daily caps. The mailbox cap stops
  without advancing the watermark, so excess waits on the server; the per-sender
  cap skips-and-advances so one flooder cannot wedge the board.

Optional DMARC enforcement reads **only** the topmost `Authentication-Results`
(hops prepend, so anything below is the sender's to forge) and cross-checks it
against the `From` domain. SPF alone does not satisfy it — SPF authenticates the
envelope, not the visible `From`.

## Verification

- **2021 unit tests**, 162 of them new, identical on PHP 8.2, 8.3 and 8.5.
- The load-bearing behaviours were proven by **mutation** — break the code, watch
  the named test fail: the UID range filter, watermark persistence on failure,
  the plain-over-HTML preference, and the e2e password-leak assertion.
- **Live check against a real Dovecot 2.3** with a CA-signed certificate, since
  the unit tests only drive a scripted fake. TLS, `LOGIN`, `EXAMINE`,
  `UID SEARCH` and a literal `UID FETCH` all behaved; an encoded-word subject and
  quoted-printable UTF-8 body survived intact. Of three seeded messages only the
  genuine one became a card — the auto-reply and the spam-flagged one were
  declined and reported. Four polls, plus one with the watermark forced to zero,
  still produced exactly one card.
- Access control checked with a real second user: a board member with edit
  permission gets `Access denied` on the whole endpoint, and the password appears
  in no response, export or backup — not even for a manager.

## Known boundaries, documented not hidden

Attachments are **named in the card body, not stored**. A reply is a **new card,
not a comment** — so this is "turn emails into cards", not helpdesk ticketing;
two-way threading needs a signed reply-address token and belongs in its own
change. A Nextcloud server admin can still recover the password from the database
plus `config.php`, the same guarantee Nextcloud gives external-storage
credentials.

Setup guide: `docs/email-intake.md`.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -326,6 +326,7 @@ Kanso 是一款为速度而生的 Nextcloud 看板应用。看板、列和卡片
 		<job>OCA\Kanso\Cron\SendDueReminders</job>
 		<job>OCA\Kanso\Cron\SendPersonalReminders</job>
 		<job>OCA\Kanso\Cron\BackupBoards</job>
+		<job>OCA\Kanso\Cron\PollMailIntake</job>
 	</background-jobs>
 	<commands>
 		<command>OCA\Kanso\Command\Rebalance</command>
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -247,6 +247,13 @@ return [
 		['name' => 'webhook#forgejoIntake', 'url' => '/api/boards/{id}/forgejo/intake', 'verb' => 'PUT'],
 		['name' => 'webhook#forgejoDisable', 'url' => '/api/boards/{id}/forgejo', 'verb' => 'DELETE'],
 
+		// Email intake (#117). No ingest route: unlike the webhooks above, mail is
+		// PULLED from IMAP by cron, so these are config-only and all MANAGE-gated.
+		['name' => 'mailIntake#show', 'url' => '/api/boards/{id}/mail-intake', 'verb' => 'GET'],
+		['name' => 'mailIntake#save', 'url' => '/api/boards/{id}/mail-intake', 'verb' => 'PUT'],
+		['name' => 'mailIntake#test', 'url' => '/api/boards/{id}/mail-intake/test', 'verb' => 'POST'],
+		['name' => 'mailIntake#destroy', 'url' => '/api/boards/{id}/mail-intake', 'verb' => 'DELETE'],
+
 		['name' => 'cardLink#index', 'url' => '/api/cards/{cardId}/links', 'verb' => 'GET'],
 		['name' => 'cardLink#create', 'url' => '/api/cards/{cardId}/links', 'verb' => 'POST'],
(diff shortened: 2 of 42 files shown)


aktasfatih:
:tada: This pull request is included in version 0.28.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.28.0).

PR #120, opened 2026-09-11
feat(attachments): show when a file was attached and record attachment changes in Activity

### LINKED ISSUE #119: Card attachments carry no visible timestamp — neither in the attachment list nor in Activity
There is currently no way to tell when a file was attached to a card. Two
separate gaps combine into that:
	1.	The attachment list renders only filename and size. The API already
returns everything needed — lib/Db/CardAttachment.php::jsonSerialize()
exposes createdAt and uploadedBy — but the card UI never renders either,
not even as a title tooltip.
	2.	Attachment changes are not written to the change log. lib/Db/Change.php
defines verbs 1–24, none of them for attachments, and
CardAttachmentService only fires ChangeNotifier for the live update
("with no new Change type" in the class docblock). So an upload never shows
up in the Activity tab either — and a removal leaves no trace at all.
Everything else on a card is timestamped: comments, checklist steps, labels,
assignees, moves. Attachments are the one exception.
Why this matters
We migrated roughly 7,300 attachments from Deck into Kanso on a production
instance. Our boards track medical-supply cases, where "when did this document
reach this case?" is a routine question — for documentation, for deadlines, and
for reconstructing who added what. Deck showed the attachment date; Kanso
currently shows it nowhere, although the value is sitting in the database
(oc_kanso_card_attachments.created_at is populated for all 7,328 rows on our
instance) and is already sent to the client on every card load.
Steps to reproduce
	1.	Open any card and attach a file.
	2.	Look at the Attachments section — only the filename and the size are
shown.
	3.	Switch to the Activity tab — the upload produced no entry.
	4.	Remove the attachment again — still no entry.
Suggested fix
a) Show the timestamp in the attachment row (frontend only).
Render createdAt next to the size, using the same formatters the activity feed
already uses — the relative form for recent uploads plus the exact date and time
as a <time datetime=…> / title, exactly as
card-modal__activity-time / card-modal__activity-exact do today. Showing
uploadedBy as an avatar or name would round it off. No API change is needed;
both fields are already in the payload.
b) Record attachment changes in the change log.
Add VERB_ATTACHMENT_ADDED and VERB_ATTACHMENT_REMOVED to Change, write them
from CardAttachmentService on upload and delete, and put the filename into the
kanso_change_details side table. Uploads and removals would then appear in
Activity with actor and timestamp like every other change.
Of the two, (b) is the more valuable half for us, because it is the only one that
also covers deletions.
Environment
	•	Kanso 0.27.0
	•	Nextcloud 34.0.3.2
	•	PHP 8.4.17
	•	MariaDB 10.11.10
	•	Ubuntu 24.04.3 LTS (LXC container on Proxmox VE)
	•	Browsers: current Safari on iOS and a desktop browser — behaviour identical

aktasfatih:
Closes #119.

An attachment carried no visible timestamp anywhere in the UI. Two separate gaps combined into that, and both are closed here.

## (a) The attachment row now answers "when?"

`CardAttachment::jsonSerialize()` already exposed `createdAt` and `uploadedBy`, but the card rendered only filename and size. The row now carries a second line: the uploader (avatar + display name) and the upload time — the same relative label + exact `<time datetime>` stamp the Activity feed uses, always rendered rather than hover-only (a tooltip is unreachable on touch, and this app is installed on phones).

The list endpoint and both upload responses now carry `uploadedByName` beside the uid, so the card credits a person instead of printing a raw uid. The uid stays in the payload — the avatar endpoint is keyed on it.

## (b) Attachment add/remove are recorded in the change log

Both paths wrote a bare `ENTITY_CARD` / `ACTION_UPDATE` row with no verb, so an upload rendered as a generic "updated this card" and a **removal left no trace at all**. Both now carry a verb (`VERB_ATTACHMENT_ADDED` = 25, `VERB_ATTACHMENT_REMOVED` = 26) plus the filename in the `kanso_change_details` side table, so the feed reads **"attached report.pdf"** / **"removed the attachment report.pdf"** with actor and timestamp, like every other card mutation.

The row stays `ENTITY_CARD` / `ACTION_UPDATE` — delta sync and the ETag key on `(entity_type, action)`, never on the verb — so realtime and polling are unchanged. No migration: the verb column and the details table both already exist.

Notes:
- "Attach from Files" records the same activity as a multipart upload; two doors into the same store must leave the same trace.
- The purge cascade (`deleteAllForCard`) deliberately writes **no** per-file "removed" activity — the card itself is being torn down and emits its own DELETE row.
- The delete path reads the filename *before* dropping the row: afterwards it is the only surviving record of which file it was.

## Verification

Run locally in a throwaway `php:8.3-cli` container (with `zip` installed) against this worktree:

- `phpunit` — **2053 tests, 919334 assertions, OK**, including 4 new cases: upload / attach-from-Files / delete each assert the verb *and* the filename detail, and the purge cascade asserts no detail row.
- `psalm` — **exit 0**, no errors.
- `php-cs-fixer --dry-run` — **0 of 417 files** need fixing.
- `npm run test:unit` — 126 pass. `npm run build` — clean. `npm run l10n:check` — extract/sync/compile re-run and the regenerated catalogues committed.

The new Playwright spec (`card-attachments.spec.js` — asserts the row's `<time datetime>` stamp, and that both the add and the remove show up in Activity naming the file) was **not** run locally: `kanso-dev` is currently mounted to a different worktree, so a local run would have silently tested other code. It runs here in CI.

### DIFF
--- a/lib/Controller/CardAttachmentController.php
+++ b/lib/Controller/CardAttachmentController.php
@@ -7,6 +7,7 @@ declare(strict_types=1);
 
 namespace OCA\Kanso\Controller;
 
+use OCA\Kanso\Db\CardAttachment;
 use OCA\Kanso\Service\CardAttachmentService;
 use OCA\Kanso\Service\NotPermittedException;
 use OCP\AppFramework\Controller;
@@ -17,6 +18,7 @@ use OCP\AppFramework\Http\DataDisplayResponse;
 use OCP\AppFramework\Http\DataDownloadResponse;
 use OCP\AppFramework\Http\JSONResponse;
 use OCP\IRequest;
+use OCP\IUserManager;
 use OCP\IUserSession;
 
 /**
@@ -32,6 +34,7 @@ class CardAttachmentController extends Controller {
 		IRequest $request,
 		private IUserSession $userSession,
 		private CardAttachmentService $attachmentService,
+		private IUserManager $userManager,
 	) {
 		parent::__construct($appName, $request);
 	}
@@ -39,9 +42,14 @@ class CardAttachmentController extends Controller {
 	#[NoAdminRequired]
 	public function index(int $cardId): JSONResponse {
 		return $this->respond(function () use ($cardId): JSONResponse {
-			return new JSONResponse(
-				$this->attachmentService->listForCard($cardId, $this->currentUserId())
-			);
+			// One display-name cache across the list: a card's attachments are
+			// usually all from the same one or two people.
+			$nameCache = [];
+			$out = [];
+			foreach ($this->attachmentService->listForCard($cardId, $this->currentUserId()) as $attachment) {
+				$out[] = $this->serialize($attachment, $nameCache);
+			}
+			return new JSONResponse($out);
 		});
 	}
 
@@ -61,9 +69,9 @@ class CardAttachmentController extends Controller {
 	public function create(int $cardId): JSONResponse {
 		return $this->respond(function () use ($cardId): JSONResponse {
 			$upload = $this->request->getUploadedFile('file');
-			return new JSONResponse(
+			return new JSONResponse($this->serialize(
 				$this->attachmentService->upload($cardId, $upload, $this->currentUserId())
-			);
+			));
 		});
 	}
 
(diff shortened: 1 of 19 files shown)


aktasfatih:
:tada: This pull request is included in version 0.29.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.29.0).

PR #121, opened 2026-09-11
Board session 2026-09-10: two leak fixes, Deck re-import guard, board freshness, ETag replay

aktasfatih:
A `/pm` board session. **11 cards, 14 commits**, every gate re-run independently before each push rather than taken from a worker's report. Branch is merged up to `main` at v0.28.0.

Board card ids below are Kanso cards (board 19), not GitHub issues.

## Security — two live holes, both found by work on something else

**`fix(security)` `2f84e17` — the trash endpoints told strangers whether a card exists.**
`DELETE /cards/{id}/purge` and `POST /cards/{id}/restore` checked *"is this card trashed?"* **before** board permission and visibility. So a live card answered `400 "Card N is not in the trash"` to **any signed-in user, non-members included**, while an unused id answered `404`. That difference is a cross-board card-existence oracle. Reordered to access-first, trash-state-last.

**`fix(security)` `7f40032` — a public board link confirmed whether any account exists.**
`PublicShareService::resolveDisplayName()` resolved *any* uid the mention pattern matched, with no membership test, on the anonymous payload. An EDIT member holding the share link could enumerate account existence and display names **instance-wide**, bypassing an admin who had deliberately disabled Nextcloud's sharee enumeration. Now member-scoped.

The trade is deliberate and recorded in the commit, the class docblock and a pinned test: a **non-member** mention now ships its raw uid, partially regressing `36523c7`. Same class already accepted for deleted accounts, and what it closes is much larger.

## Adoption

**`fix(import)` `0ca0005` — re-running a Deck import no longer duplicates the board and every attachment's bytes.**
The importer had no memory of what it had imported (`grep deck_board_id` → zero hits) and the picker offered an unconditional Import button. A successful-but-lost response looked identical to a real failure, so retrying was the obvious move — and produced a second complete board plus a second physical copy of every attachment.

Now recorded per **(Deck board, importing user)** in `kanso_deck_imports`; the picker shows "Imported &lt;date&gt;" and puts a second copy behind a confirmation that states plainly that nothing from the earlier import is reused.

The **unique index is the guard, not the pre-check** — the ledger row is the transaction's first write, so a concurrent double-submit loses at the database *before any attachment byte is copied*.

## Performance and correctness

**`perf(board)` `6373057`** — the board read now replays its ETag. Measured, median of 40, unchanged board: **123 887 B → 1 512 B (−98.8%)** on 180 cards, **1 319 002 B → 1 513 B (−99.9%)** on 2 000. A changed board still returns 200 with no penalty.

**`fix(board)` `e5fca37`** — three things that compounded. The missing `kanso_changes` writers (`AutomationService` zero-second branch, `RecurrenceService` CRUD, `DueReminderService`) now log; an empty delta tick no longer refreshes `dataUpdatedAt`, which had made `refetchInterval`/`refetchOnMount`/`refetchOnWindowFocus` **all dead while a board was open** (measured: zero `/boards/{id}` reads in 100 s); and the ETag became **viewer-aware** (`<change id>-<mask>-<role>`), because the payload is viewer-scoped while the validator was not — so a partial group revocation used to 304 straight past. Costs **+11 bytes and +0.5 ms** against the commit above.

**`fix(board)` `8a4dea6`** — a board whose access is revoked no longer keeps rendering from cached data. Only **403/404** is terminal; a 500 or a dropped connection still shows the cached board, and both halves are asserted.

## UI, CI, docs, test integrity

- **`fix(ui)` `d82cb49`** — the bulk action bar was **502 px wide at a 360 px viewport**, `position: fixed`, so the end controls were clipped off-screen and unreachable on a phone. Now an overflow menu **with visible text labels**, plus the `NcActions` accessible-name fix (a screen reader used to say "Bug" for the add-label button).
- **`ci` `df89357`** — dependabot PRs skip the browser suite. Implemented so the job still **reports** under the name `e2e`: it is a *required* check, and a job-level `if:` would leave it never reporting and permanently block every dependabot PR from merging.
- **`test(e2e)` `ed80377`** — the visibility leak matrix now sweeps its whole card-addressed route table with a non-member viewer, turning a three-route fix into a suite-wide "membership before visibility" invariant.
- **`test(e2e)` `7c91270`, `077b313`** — dropped-keypress guards, and three specs that asserted on unfiltered console output and tripped on other Nextcloud apps' noise.
- **`docs` `07ff0d4`** — README records that offline PWA data outlives a logout on a shared device.
- **`fix(migration)` `b56863f`** — renumbered the Deck-import migration to `Version006200` after the email-intake merge took `006000` and `006100`.

## Verification

**2076 PHPUnit · 141 frontend unit · psalm 0 · php-cs-fixer clean · build, info-xml, l10n:lint, l10n:check all green.** Fresh installs verified on **postgres, sqlite and mariadb**, and the migration renumber re-proven by a clean install carrying both this branch's and email intake's tables.

Every load-bearing behaviour was **proven by mutation** — break the code, watch the named test fail, restore. Applied per route and per half, never batched. Two cases where that earned its keep: a mutant that was **green until a call-count assertion was added**, and a CDP CPU-throttle "proof" that was **discarded as false** because it left the old code passing.

## Worth a reviewer's attention

Every card validated in grooming was **materially wrong**, and **eight findings came out of execution with no card at all** — including an install-breaking primary-key name for NC 30–32 (caught by the repo's own `PrimaryKeyNameLengthTest`), a MySQL lock-wait path returning 500 instead of 409 in exactly the double-submit case the feature exists for, and a pre-existing test that had **frozen a bug in place** by asserting the buggy behaviour was correct.

No `appinfo/info.xml` `<version>` bump — release-only, per the charter.


aktasfatih:
:tada: This pull request is included in version 0.30.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.30.0).

PR #123, opened 2026-09-11
feat: sub-card dates in the parent list, and the column on every search hit

### LINKED ISSUE #122: Sub-card dates on the parent card, and the column in search results
Two small display wishes from daily use. Happy to split them into separate issues if that's easier to track.

1. Show start/due date of sub-cards on the parent

Right now the sub-card list shows a status dot, the title and a detach button. To see when a sub-card starts or is due, I have to open it.

We use a parent card for a case and sub-cards for its steps, each with its own dates. Seeing those dates in the sub-card list would save a lot of clicking.

From a quick look, children[] in the card payload already comes from Card::jsonSerializeSummary(), which includes duedate and startDate — so this might be display-only. Something compact like the date chips on a card tile would already do the job.

2. Search results don't show the column

A search hit shows the icon, card title, snippet and type badge, but not which column the card is in.

In our boards a customer often has several open cases with very similar titles — the column is what tells them apart, since it reflects the stage the case is at. So the result list often has several rows that look the same and I have to open them one by one. Deck used to show the stack, which is why I noticed.

As far as I can tell the column isn't part of the search response at all (SearchService returns type, cardId, boardId, title, snippet, rank), so this one probably needs a backend change too. CardMapper::fetchAssignedRows() already returns a stackTitle for the "My tasks" feed, if that helps as a starting point.

Environment
Kanso 0.27.0, Nextcloud 34.0.3.2, PHP 8.4.17, MariaDB 10.11.10
Ubuntu 24.04.3 LTS (LXC container on Proxmox VE)

Saygilar 

Fatih Ekici

aktasfatih:
Closes #122 — both asks, one commit each so the changelog gets both entries.

## 1. Sub-card dates, and a sub-card list that reads like a list

`children[]` already carried `duedate`, `startDate` and `allDay` from `jsonSerializeSummary()`; the UI rendered none of it, so seeing when a step starts or is due meant opening it.

Three changes:

- **One sub-card per row.** It was a two-column grid, which read as a matrix rather than an ordered set of steps, left a ragged hole on an odd count, and gave each cell too little width to carry more than a title.
- **Dates on each row** — a quiet start chip and a filled due chip, with the board tile's overdue / due-soon colouring, including its rule that a *done* sub-card shows its due date neutrally instead of shouting red about finished work. No dates → no chips. `allDay` is honoured, so an all-day date no longer renders as the previous day west of UTC.
- **Mobile**: the dates drop to a second line, indented under the title and clear of the detach button, rather than crushing the title to an ellipsis. Same four elements re-flowed by grid areas — no duplicated markup.

### Legibility fix this surfaced

NC 34's `--color-error` / `--color-warning` are pale **tint backgrounds** (`#FFE7E7` / `#FFEEC5`), not saturated fills — so the `color: #fff` the overdue and due-soon chips carried put white on near-white. Measured:

| chip | before | after |
|---|---|---|
| overdue | **1.18:1** | **7.92:1** |
| due soon | **1.15:1** | **6.78:1** |

Both now derive text and background from the same theme-aware `--color-*-text` token, so dark mode is handled upstream rather than by a hardcoded override. This also fixes the **existing** checklist-step due chips, which had the identical bug.

## 2. The column on every search hit

`SearchService` emitted `type, cardId, boardId, title, snippet, rank` — no column. The reporter's boards hold several open cases per customer with near-identical titles where the column *is* the distinguishing detail, so result rows looked alike and had to be opened one at a time.

- Card hits already carried `stackId`; `CommentMapper::searchInBoards()` now selects `c.stack_id` too.
- New `StackMapper::titlesByIds()` resolves titles **across boards in one query** — the existing `findByIds()` is board-scoped and search spans every readable board. Deliberately not filtered on archived: archiving a column doesn't archive its cards, so an archived column still holds live, searchable cards.
- Resolved for the returned **page**, after slicing — a wide search can match hundreds of rows, and all but one page is about to be discarded. An unresolvable stack yields a null title and the client renders no chip.
- Shown in the search box and the command palette, sharing a line with the comment badge so no hit grows a third row.

## Verification

Against a local NC 34 stack running this branch:

- **phpunit** — 2056 tests, 919346 assertions, **OK** (5 new: every-hit-names-its-column, unresolvable stack → null, page-only resolution, plus a stale comment fixture corrected to the mapper's real shape).
- **psalm** exit 0 · **php-cs-fixer** 0 of 417 · **npm run test:unit** 126 pass · **build** clean · **l10n:check** regenerated and committed.
- **e2e** — `search.spec.js`, `command-palette.spec.js`, `parent-child.spec.js`: **20/20 green**, including a new spec that seeds two identically-titled cards in different columns and asserts the column is what separates them. Proven non-vacuous: nulling `stackTitle` server-side fails it deterministically (3/3 retries), reverting restores green.
- Sub-card layout checked at **1400px and 390px**; no row overflows its container on mobile, and chip variants were asserted per row (overdue / soon / start-only / done-neutral / no-dates).

### DIFF
--- a/lib/Db/CommentMapper.php
+++ b/lib/Db/CommentMapper.php
@@ -222,8 +222,8 @@ class CommentMapper extends QBMapper {
 				'cardId' => (int)$row['card_id'],
 				'boardId' => (int)$row['board_id'],
 				// The card's column, so a search hit can name the stage it is at
-				// (#122). Carried as an id; SearchService resolves the titles for
-				// the whole result set in one query.
+				// (#122). Carried as an id; SearchService batch-resolves the titles
+				// for the page it returns.
 				'stackId' => (int)$row['stack_id'],
 				'cardTitle' => (string)$row['title'],
 				'body' => (string)$row['body'],
--- a/lib/Db/CommentMapper.php
+++ b/lib/Db/CommentMapper.php
@@ -192,17 +192,18 @@ class CommentMapper extends QBMapper {
 	 * Non-deleted comments whose body matches a LIKE pattern, restricted to the
 	 * given readable boards (joined through non-deleted cards). Portable
 	 * case-insensitive LIKE; the pattern is pre-escaped/wrapped by the caller.
-	 * Each row carries the parent card's id, board and title so a hit can be
-	 * shown and deep-linked without a second query. $boardIds must be non-empty.
+	 * Each row carries the parent card's id, board, stack and title so a hit can
+	 * be shown and deep-linked without a second query. $boardIds must be
+	 * non-empty.
 	 *
 	 * @param int[] $boardIds
 	 * @param array<int, string> $rolesByBoard the viewer's role per board id
-	 * @return array<int, array{id: int, cardId: int, boardId: int, cardTitle: string, body: string}>
+	 * @return array<int, array{id: int, cardId: int, boardId: int, stackId: int, cardTitle: string, body: string}>
 	 * @throws Exception
 	 */
 	public function searchInBoards(array $boardIds, string $likePattern, int $limit, string $uid, array $rolesByBoard): array {
 		$qb = $this->db->getQueryBuilder();
-		$qb->select('cm.id', 'cm.card_id', 'cm.body', 'c.board_id', 'c.title')
+		$qb->select('cm.id', 'cm.card_id', 'cm.body', 'c.board_id', 'c.stack_id', 'c.title')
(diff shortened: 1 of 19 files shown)


aktasfatih:
:tada: This pull request is included in version 0.30.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.30.0).

Issue #127, opened 2026-09-12
Feature Request - Archive all cards in this list


impronta48:
This is a feature I use a lot, much easier than "select multi - select all - archive all"

<img width="687" height="619" alt="Image" src="https://github.com/user-attachments/assets/edca194e-452d-4fbe-b517-cd59f04da357" />

aktasfatih:
:tada: This issue is fixed in version 0.31.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.31.0).

If Kanso saves you time, consider [sponsoring its development](https://github.com/sponsors/aktasfatih) — it’s an AGPL solo project and sponsorships keep it maintained. 💜

Your semantic-release bot :package::rocket:

Issue #126, opened 2026-09-12
When choosing a collaborator the tooltip overlaps


impronta48:
the name of the selected person is under the tooltip

<img width="573" height="664" alt="Image" src="https://github.com/user-attachments/assets/cd416f03-489d-47e1-9512-19cbe5739bd3" />

aktasfatih:
:tada: This issue is fixed in version 0.31.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.31.0).

If Kanso saves you time, consider [sponsoring its development](https://github.com/sponsors/aktasfatih) — it’s an AGPL solo project and sponsorships keep it maintained. 💜

Your semantic-release bot :package::rocket:

PR #128, opened 2026-09-13
feat: address three reported issues, plus the bugs found fixing them

### LINKED ISSUE #125: Page Title should follow Board title
It would be great that the page title would follow the board Title like

Personal | Kanso | Nextcloud

This is useful when you have bookmarks of your boards for quick access

aktasfatih:
## Reported by users

- **Browser tab title now names the board / card / list you're on** (#125). There was no title machinery at all — the tab read `Kanso - Nextcloud` on every route. New `usePageTitle` composable captures Nextcloud's own server-rendered title once at module load and maintains a stack of claims, wired into 14 views. The suffix is captured rather than hardcoded (the app name is translatable and the instance name is themable), and the separator is core's ` - `.
- **The avatar tooltip no longer covers the collaborator's name** (#126). Not a z-index bug: `NcAvatar`'s tooltip in @nextcloud/vue 9 is a plain native `title=`, anchored at the cursor, duplicating the name 8px away. Flipped `disable-tooltip` on the four avatars that sit beside their own name and moved the title onto the ellipsized name span — which also makes a truncated name readable for the first time.
- **Archive every card in a column from one menu entry** (#127). Reuses the existing bulk endpoint; the backend delta is one enum value (`unarchive`) plus a short case, which also makes the pre-existing bulk-archive-selected undoable.

## Bugs found while doing the above

- **A selection of more than 100 cards failed every bulk action** with a 400. `MAX_CARDS = 100` is a hard server cap but the client never chunked, and shift-range select has no limit. One chunker now serves both the selection actions and the column actions.
- **Changing any setting silently cleared your "default board on start."** `PUT /api/settings` wrote `default_board` unconditionally while each toggle sends only its own key. Reproduced live (`440` → `null`). Fixed with an `array_key_exists` check on the raw body — deliberately not the repo's existing `getParam` sentinel, which is `isset`-based and would have broken the legitimate "clear my default board" action.
- **A deleted board was never actually deleted.** `deletedAt` was a tombstone with no purge and no restore, so rows and attachment bytes lived forever unreachable. Now a declarative cascade over the board-scoped tables (derived by scanning all 62 migrations, with a test that fails when a new table goes unclassified), bytes swept outside the transaction, reaped by a batched cron job after 30 days.
- **Pruning the change log orphaned its detail rows instance-wide, forever.** `change_details` has no `board_id` and no `card_id`, so once its parent was pruned nothing could ever reach it. Child-first delete going forward, plus a bounded one-off background job that drains the existing backlog and then stops.
- **Board statistics showed the bare app name and "Stack 12" bars** on a hard load, because it peeked the query cache non-reactively instead of running a board query.
- **An auto-archive rule could not be widened back to the whole board**, despite the API's own docblock promising it — same `isset`-based sentinel root cause.

## Docs and infrastructure

- **The README was telling people Kanso isn't on the App Store** (it has been since August) and that **the tarball isn't signed** (it is). Installation now leads with the App Store; the language list is derived from `l10n/` behind a CI drift guard; a dozen shipped-but-undocumented features (attachments, time tracking, custom fields, CalDAV, Trello/CSV import, public sharing, reminders, backups, templates, saved filters, swimlanes) are now listed — each verified reachable, with half-built sub-paths deliberately left unclaimed.
- **A static landing page** plus a Pages deploy workflow. **Pages is not enabled** — that's a repo setting, left for a human: Settings → Pages → Source → GitHub Actions.
- **The dev stack's `/push` proxy survives container recreation.** It was configured by `docker exec` into the running container, so any `docker compose up -d` silently wiped it while Nextcloud kept advertising the push capability — advertised-but-dead, which red-lined six unrelated e2e specs locally.

## Verification

Every card was implemented by a worker that ran the gates itself and **proved its tests non-vacuous by mutation** — breaking the fix, confirming the test goes red, restoring. That caught real problems: the archive-rule tests were vacuous (they stubbed `getParam` to return `null` for the "key present" case, which the real accessor never does), and one worker's inherited diff had a mutation edit left unreverted that would have shipped the bug intact under a comment claiming otherwise.

PHPUnit 2093 → 2136. psalm and php-cs-fixer clean throughout. No `<version>` bump — the migrations land unbumped per the release rule.

Closes #125, #126, #127


aktasfatih:
:tada: This pull request is included in version 0.31.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.31.0).

PR #133, opened 2026-09-14
fix: correctness, accessibility and store-listing follow-ups

aktasfatih:
A sprint of eight follow-ups from the last one — correctness, accessibility, and the store listing. No user-reported issues in this batch; these are things the previous sprint's work surfaced, plus one long-open decision.

## Correctness

- **A bulk archive now reports only the cards it actually moved.** Previously every card landed in `ok` whether or not it transitioned, so a card someone *else* archived inside the stale-cache window rode along and got **unarchived by your undo**. The per-card op contract is now `callable(int): bool`; a no-op lands in `skipped` with reason `unchanged`. Response shape unchanged, so no frontend edit was needed.
- **The column action stops calling itself "Archive all cards."** It archives what you can see — correct behaviour, but "all" was untrue. It now names the count. Deliberately *not* fixed by reporting how many hidden cards were left behind: `LeakMatrixTest` pins that a count is itself a leak surface, and a hidden-card count is exactly the existence oracle card visibility exists to deny. That reasoning is recorded in code so it isn't re-proposed.

## Storage

- **An admin can cap total attachment storage for the instance** (`occ config:app:set kanso attachment_storage_limit`). **Off by default** — with no value set the behaviour is byte-identical to today, and no aggregate query runs at all. Attachment bytes live in app-owned storage *outside* the Nextcloud quota, bounded only at 100 MiB per file, so any authenticated user could previously loop uploads until the host disk filled and took the whole instance down. Enforced on every unbounded write path including Deck import, which turned out to be a third writer with no byte bound and no rate limit.

## Accessibility and polish

- **Reduced motion is honoured by the three JavaScript scrolls** that were bypassing it. The app already had a CSS guard, but per CSSOM-View an explicit `behavior` option overrides the CSS property, so it never reached them.
- **A truncated collaborator name can be read on hover** in seven more places — completing the pattern an earlier fix established.
- **A card link paints your chosen layout immediately** instead of rendering the default and snapping. Settings now ship as initial state from both entry points; this also fixes the older editor-toolbar flash.
- **A public board link has a tab title of its own.** Every public share link on an instance previously shared one identical title, because the public page renders outside the branch that sets one.

## Store listing

- **The App Store description was audited and all ten translated blocks re-synced.** The listing mentioned only Deck import and omitted a dozen shipped features — attachments, custom fields, time tracking, reminders, public links, CalDAV, email intake, Trello/CSV import among them. Every translated block had also drifted (all ten still said GitHub where the English says GitHub and Forgejo). `info.xml` sits outside the `.pot` pipeline entirely, which is how that happened; `docs/TRANSLATING.md` now records it.

## Verification

Every card was implemented by a worker that ran the gates itself and **proved its tests non-vacuous by mutation**. That caught real problems in this batch:

- The public-share fix's originally-specified scope would have rendered **"My Board - My Board - Nextcloud"** — the client-side and server-side titles are mutually exclusive, not additive.
- The first no-wait first-paint test was **still vacuous**: the card modal enters the DOM after the settings request has already answered locally, so the mutation passed. Rewritten to capture the class list at first paint and hold the request until then.
- The archive-label change's negative ACL assertion was re-proven able to fail before being trusted.
- The attachment cap's own code review found Deck import as an uncapped third writer, which would have made the documented claim false.

PHPUnit 2136 → 2160. psalm and php-cs-fixer clean. No `<version>` bump.

**Two things worth a human eye before merge:**
1. The **Polish, Russian and Chinese** description blocks are faithful in meaning but were written without a native speaker; the new integrations paragraph could use smoothing. This is public store copy.
2. The attachment cap shipped on a decision inferred from two review approvals rather than an explicit answer. It is off by default and reversible, but say the word and it comes out.


aktasfatih:
:tada: This pull request is included in version 0.32.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.32.0).

PR #143, opened 2026-09-14
test(e2e): locate toasts by ARIA role and message, not @nextcloud/dialogs classes

aktasfatih:
Unblocks Dependabot PR #124, which has been red since 2026-09-12 on a single check.

## What was wrong

Five e2e specs asserted on **Toastify's internal class names** — `.toastify`, `.toast-success`, `.toast-warning`, `.toast-undo`. Those are an implementation detail of `@nextcloud/dialogs`, not a contract.

PR #124 bumps `@nextcloud/dialogs` 7.4.1 → 7.5.0, which **drops `toastify-js` from its dependency tree entirely** and rebuilds toasts on `@nextcloud/vue`. The classes vanish, and those assertions fail with "element(s) not found" — deterministically, on every retry.

**This was never a product regression.** `src/` has zero references to those classes; Kanso only ever calls `showSuccess` / `showWarning` / `showUndo` and lets the library render. Toasts work fine for users. Only the tests reached into library internals, and nothing here changes application code.

## The fix

A thin `toast(page, text)` locator in `tests/e2e/helpers.js`, matching a toast as what it actually is to a user — an **ARIA live region** carrying a message:

- `role="status"` / `role="alert"` — what dialogs ≥7.5 renders
- `aria-live` on the toast element — what ≤7.4 renders
- `:not(.hidden-visually)` — excludes the screen-reader-only announcement regions 7.5 mounts alongside every toast
- `.filter({ hasText })` on the user-visible message

Undo buttons now use `getByRole('button', { name: 'Undo' })` rather than a class filter.

That exclusion is load-bearing, not cosmetic: without it, 7.5.0 produces a strict-mode violation resolving to two elements — and a spec could pass on the invisible announcement region while no toast was on screen at all.

## Verification

Run against **both** dependency versions, because passing on only one would defeat the purpose:

| | 7.4.1 (current `main`) | 7.5.0 (what #124 installs) |
|---|---|---|
| Five spec files | **18/18 passed** | **18/18 passed** |

The 7.4.1 pre-change baseline was also 18/18, so there is no behaviour drift — only decoupling.

**Proof the 7.5.0 bundle was genuinely live:** with the helper temporarily set back to `.toastify`, copy-as-prompt failed with "element(s) not found" — exactly the breakage #124 hits today.

**Mutation proof, both directions** — a toast locator can fail by matching nothing *or* by matching everything:

- *Matches nothing:* helper set to a nonexistent role → all five toast tests failed, on both versions.
- *Matches the wrong thing:* patched the source to raise a **warning with different text** instead of the expected success toast → copy-as-prompt failed on both 7.4.1 and 7.5.0. A real toast of the wrong type and message does not satisfy the assertion. Source mutation reverted.

`package.json` and `package-lock.json` are **unchanged** — 7.5.0 was installed with `--no-save` for the verification run and `npm ci` restored 7.4.1 afterwards. The commit touches six test files only.

`npm run test:unit` 152/152, `npm run build` clean, `l10n:check` no diff.

## After this merges

**#124 needs a rebase** to pick up the new selectors — it isn't touched by this commit. Its other 14 checks were already green.


aktasfatih:
:tada: This pull request is included in version 0.32.1 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.32.1).

PR #145, opened 2026-09-15
fix: the three accepted issues — list rendering, public-share card sections, light-mode contrast

### LINKED ISSUE #135: Checklist items & subcard references not displayed in public share
### What happened?

Checklist items and subcard references are not displayed in a public share when opening a card.

### Steps to reproduce

Create a card with checklist items and a subcard. Share the board public.

Image

Image

### Kanso version

0.31.0

### Nextcloud version

34.0.4

### Database

None

### Errors

```shell

```

### Before submitting

### LINKED ISSUE #136: Light mode support
### What's the problem?

Improve the color design for Nextcloud in light mode for priority / type / date / ...
Image

### What would you like it to do instead?

_No response_

### How much does this matter to you?

Nice to have — I'd use it, but I'm fine without it

### Anything else?

_No response_

### Before submitting

- [x] I searched the existing issues and this isn't already requested

### LINKED ISSUE #139: List formatting incorrectly / not displayed outside of edit mode in card description
### What happened?

The formatting of an unordered list is not displayed outside of the edit mode in a card description.
Edit mode:
Image

After saving:
Image

The formatting of an ordered list is displayed incorrectly outside of the edit mode (left-alignment is wrong) in a card description.
Image

### Kanso version

0.31.0

### Nextcloud version

34.0.4

### Before submitting

aktasfatih:
## `fix:` show bullets and indentation on lists in saved card descriptions — #139

Lists rendered correctly in the editor but lost their bullets (and, for ordered lists, their indent) once saved. The markdown pipeline was never at fault: `markdown-it` emits proper `<ul>/<ol>/<li>` and DOMPurify allowlists them. Nextcloud's `core/css/server.css` carries an unscoped `ul { list-style: none }` plus a Meyer-style reset that zeroes list padding, and the display container — unlike the editor — re-declared nothing.

Adds a scoped `.kanso-md` class (`src/styles/markdown.css`) restoring markers, `padding-inline-start` (logical, so RTL works), nested marker styles, blockquote borders and `h1` sizing, applied to all nine rendered-markdown containers. The public share page gets the equivalent inline, since its bundle imports no CSS. The NC reset is left untouched globally — relaxing it would put bullets inside `NcActions`/`NcAppNavigation` internals.

## `fix:` show checklist items and sub-cards when opening a card on a public board link — #135

The public share has no per-card endpoint, so everything the read-only modal can render must already be in the one board payload. That payload carried checklist **counts** only and no parent/child linkage at all.

Adds `checklistItems` (a hand-written `{title, done}` shape — deliberately **not** `ChecklistItem::jsonSerialize()`, which carries `assignedUser` and `assignedRole`) and `childIds`, derived from the already-filtered public card set rather than `findChildren()`/`findVisibleChildren()`, both of which bypass the public-only visibility scope. Checklist items come from one new grouped query carrying the same `applyPublicOnly` scope as the existing count query, so the endpoint keeps its constant query count; the sub-card edges cost zero extra queries.

Both key-set leak guards were updated in their two sources of truth, a `PUBLIC_CHECKLIST_ITEM_KEYS` guard added, and `assigned` added to the forbidden-substring sweep — `assignee` did not cover `assignedUser`.

## `fix:` make priority, type, due-date and error colours readable in light mode — #136

Nextcloud 32 redefined `--color-error`/`--color-warning`/`--color-success` from *foreground colours* into *background tints* (`#DB0606` → `#FFE7E7`), moving the foreground role to `--color-*-text`. Kanso still used the pre-32 semantics, so in light theme the affected text rendered at **≈1.16–1.18:1 — effectively invisible**. This reached well past the badges in the report: the same token drives error messages across the app, and the label-delete button was `#fff` on `#FFE7E7`.

A second defect sat behind it: the dark-mode overrides keyed on `body.theme--dark` / `body.theme--default` / `body:not(.theme--light):not(.theme--dark)`, but Nextcloud emits `data-theme-<id>` and `data-themes`. Two selectors were dead and the third matched unconditionally, so a user on the explicit light theme whose OS prefers dark received dark colours on white — all failing AA. That block was copy-pasted in eight files.

Consolidates the four `--kanso-*-legible` tokens into one `src/styles/status-tokens.css` resolved off the theme-aware `--color-*-text` pair, deletes all eight duplicated definition and override blocks, retires the stale `-rgb` companions in favour of `color-mix`, and repoints 55 foreground declarations across 21 files plus the surface-coloured sites. Measured: **1.16:1 → 7.68:1** (priority badge), **1.18:1 → 10.09:1** (type pill, label error). Dark mode verified not to regress.

Deliberately deferred, with reasons recorded on the board: stats bar fills, the provider-mirroring link-state badges, filter dots, and the decorative gradient presets — all opaque, theme-invariant, and meeting the 3:1 graphical-object rule.

---

### Verification

Every fix carries a test proven to fail without it, not merely written alongside it:

- **#139** — new list test goes `Expected "disc" / Received "none"` with the CSS stubbed out, deterministically across all retries.
- **#135** — three separate mutations each turned a guard red and were reverted: child ids sourced unfiltered (private child leaked, 3 ids), `applyPublicOnly` dropped from the new mapper query, and `checklistItems` delegated to the serializer (key-set + substring guards both fired).
- **#136** — the new contrast spec is red pre-fix at the real numbers above and green post-fix in both themes.

Suites run: `phpunit` full (2165 tests), psalm, php-cs-fixer (0 of 429), and the `markdown`, `public-share`, `visibility-leak-matrix`, `light-contrast`, `card-type`, `checklist`, `labels`, `priority`, `review-types`, `visual`, `quick-preview`, `projects`, `comments` and `keyboard-shortcuts-acl` e2e specs. `npm run l10n:check` clean. The full e2e suite runs here in CI.

### Notes for review

- The light-mode pass is large but mechanical — one substitution rule (*foreground use → `--color-*-text`; surface use → explicit legible foreground*), not ~90 independent design decisions.
- Two incidental findings, **not** addressed here: `.label-settings__error` is unreachable via an empty or duplicate label rename (`saveRename()` returns early on empty, and `LabelService` has no uniqueness constraint) — whether duplicate label names *should* be rejected is a product question; and the board-settings error text was broken in **both** themes at 1.39:1, since that component never had a dark override.
- No `<version>` bump — versioning is release-only here.

Closes #139
Closes #135
Closes #136


aktasfatih:
:tada: This pull request is included in version 0.32.1 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.32.1).

PR #148, opened 2026-09-15
Accepted bug fixes, write-path rate limits and column select-all

### LINKED ISSUE #146: Closing a card while editing its description blocks the board
### What happened?

When closing a card using the X / close button while the card description is still being edited, the card closes immediately. A confirmation dialog then appears warning that unsaved changes will be lost.

If Abort/Cancel is selected in this dialog, the card remains closed, but the board becomes unresponsive to mouse clicks. In particular, clicking cards no longer opens them.

The edited description is also lost.

### Expected behavior
When closing a card with unsaved description changes:
- The card should remain open while the confirmation dialog is displayed.
- Abort/Cancel should dismiss the confirmation dialog and return to the card with the edits intact.
- If the user confirms discarding the changes, the card should close normally.

### Steps to reproduce

1. Open a card on the board.
2. Click into the card description and make a change.
3. Without saving the description, click the X / close button.
4. A confirmation dialog appears warning that the changes will be lost.
5. Click Abort/Cancel.
6. Try to click another card on the board.

Image

### Kanso version

0.32.1

### Nextcloud version

34.0.4

### Database

None

### Errors

```shell

```

### Before submitting

### LINKED ISSUE #147: Large images in card descriptions overflow the description width
### What happened?

Large images inserted into a card description do not respect the width of the description area when the card is viewed outside of edit mode.

The image is displayed at its original width, causing it to overflow the card description area instead of being constrained to the available width.

Images in card descriptions should be responsive and constrained to the width of the description box.

For example, the image should scale down when necessary so that:
- Its maximum width is the width of the description area.
- Its aspect ratio is preserved.
- Smaller images are not unnecessarily enlarged.

The image width, etc. is handled correctly while editing the description.

### Steps to reproduce

Edit mode:
Image

After saving:

Image

### Kanso version

0.32.1

### Nextcloud version

34.0.4

### Database

None

### Errors

```shell

```

### Before submitting

aktasfatih:
Eleven cards from the 2026-09-15 board session: the two accepted user-reported bugs, four pieces of write-path and feed hardening, three half-shipped features finished, and two interaction fixes.

## User-reported bugs

**`fix(card)` — closing a card with the X while editing no longer freezes the board** (#146)
The reported "board goes dead" was not a stuck overlay. `CardModal` passed `:show="true"` with no `update:show` listener, so NcModal's close wrote into `useModel`'s *local* value — `props.show` never changed, the restoring `watchSyncEffect` never fired, and the dialog was hidden permanently. Because `BoardView` renders an unkeyed `<router-view>`, the next card reused that same hidden instance: clicking a card did navigate, it just rendered nothing.

This also surfaced a second bug shipping since 0.32.1: NcModal's Escape hotkey reaches the shell in the capture phase, *before* `CardDetail`'s root handler, so Escape with an attribute popover open closed the whole card ~300 ms later. `card-modal-layout.spec.js:310` was only passing by racing that timer.

**`fix(card)` — large images stay inside the card description** (#147)
Rendered markdown had no `img` rule anywhere; the only one in the app was the editor's own ProseMirror block — exactly why width was right while editing and wrong after saving. A shared `.kanso-md` stylesheet already existed with the class already on every rendered container, so one rule covers the description, comments, quick preview and project view.

## Hardening

**`perf(views)` — the Views feed loads in a fixed number of queries.** `ViewService` was the only cross-board feed looping `contextFor()` per board, contradicting the batching contract written on `BoardAccess` itself — and refetching every 60s. Now one `rolesFor()` plus board-set-scoped mapper twins. **Measured 96 → 18 queries at 6 boards and 287 → 17 at 19**; the per-board slope is gone. Output proven byte-identical across 27 captures (3 accounts × 9 filter shapes), with the fixture built for mixed roles — owner, internal, external and non-member — so the visibility masking is exercised, not assumed. The four viewer-scoped mappers got standalone twins rather than delegating, leaving the board-scoped SQL behind `BoardController::show()` untouched.

**`fix(import)` / `fix(board)` — rate limits on the last unguarded write paths.** Deck import had none; validation found Trello equally unguarded. Deck gets 120/3600 rather than the siblings' 60 because it is one board *per request* against a server-enumerated ACL-visible list — stranding a half-finished migration is the worse failure. Trello stays at 60. Then `BoardPortabilityController::duplicate()`, the last one, and the easiest to trigger: no client-supplied document at all, just a board id the caller can already read, cloning the whole card graph. Read paths stay unlimited, per the convention pinned by `testReadPathsAreNotRateLimited()`.

## Half-shipped features, finished

**`feat(sharing)` — set an expiry date on a public board link.** `expiresAt` was persisted and enforced but nothing could ever set it: dead config with live enforcement that had never executed. Now a MANAGE-gated endpoint and a date picker. An expired link keeps the **same 404 status and the same throttle** as every other rejection — only the message differs — and the machine-readable route stays a uniform `not_found`. A security review caught a real fail-open (a malformed `expiresAt` cast to 0 read as "clear the expiry"); the client now distinguishes empty from unreadable and the controller takes `mixed` so the framework's int cast can't launder garbage into a clear.

**`feat(attachments)` — attach a file you already have in Files.** The endpoint and a frontend helper existed; nothing reached them from inside a card. The backend **copies** rather than links (`fopen` → `newFile` under a server-generated key, no source fileid persisted), so the UI says so: the picker confirms with "Attach a copy" and a persistent hint explains that editing or deleting the original won't change it. The e2e proves the semantics by deleting the source file and re-downloading.

**`fix(board)` — card templates while grouped into swimlanes.** `SwimlaneRow` rendered the column menu but didn't forward the template props, so grouping a board silently removed templates with no indication why. Fixed, and the card asked for an item-by-item audit of all 29 props rather than just adding the missing two — which turned up a second live bug (`newCardsOnTop` wasn't forwarded, so multi-line paste into a lane composer created cards in reverse order). The 13 props a lane deliberately doesn't get now carry a comment saying why.

## Interaction and polish

**`feat(board)` — select every card in a column from the column menu.** Mirrors the shipped archive-all pattern: filter-aware label and count, additive union, hidden on empty columns. Cards outside the virtualized window are included — the test seeds 45 cards, asserts fewer are mounted, then asserts all 45 select.

**`fix(board-settings)` — the board name field uses the whole row in normal text.** `.board-settings__name-input` was in the markup but defined nowhere, so the name silently inherited the Card-ID-prefix field: a fixed 120px box, forced uppercase and monospace — about 12 visible characters for a `maxlength="100"` field. Measured 120px → 217px with normal casing; the prefix field verified unchanged.

**`fix(import)` — the Import menu says what each entry does.** CSV was captioned "Import a board" when it adds cards *into* one; it now sits under its own caption. The disabled "GitHub Projects (coming soon)" entry, which had nothing behind it and no ship date, is removed.

## Verification

Every card was implemented and tested against a dev stack confirmed to be mounting this branch. **Each new assertion was proven non-vacuous by mutation** — the fix reverted, the test observed failing, then restored; the Views card alone ran 9 mutations including authz ones in real SQL, after review caught that its PHPUnit tests mock the mappers and so never exercised the new call sites. The full Playwright suite was deliberately left to CI.

Three things for review, none changed here:
- The three `NcCheckboxRadioSwitch` instances under board settings → General omit `type="switch"` and render as checkboxes while the same component is a toggle elsewhere in that modal. One word to change, but a visual decision.
- Multi-select remains flat-board only — swimlane and List views never passed selection props down.
- `BoardListView` gets `onFetchTemplates`/`onCreateFromTemplate` but no `onManageTemplates`, so list view has the template picker without the manager entry.

## Later in the session

**`feat(board)` — edit a saved auto-archive rule.** A rule's scope, condition and threshold were fixed at creation; changing any meant delete + recreate, while the rule list displayed a scope you couldn't edit. #10461 had already fixed the backend so a rule can be re-scoped to the whole board, but nothing in the UI reached it. The Edit button loads the rule into the existing create form, mirroring the recur-rules section next door rather than adding a second set of controls.

**`fix(board)` — clear leftover notifications when a deleted board is permanently removed.** The purge swept only `kanso_` tables, leaving notification rows that no handler ever deletes and that keep inflating the unread badge. Now swept via `IManager::markProcessed()` — no direct core-table write — scoped to card **and** checklist-item ids (`step_assigned` notifications are keyed by the latter). The Activity stream has no supported per-object delete, so that residue is documented in the purge service with the evidence rather than silently skipped. A mutation test guards the dangerous direction: widening the sweep to a foreign card id fails, so it cannot delete another board's notifications.

**`fix(board)` — reach the card template manager from list view**, and reach card templates at all while grouped into swimlanes. Two omissions in hand-maintained prop lists; both binding sites now document every prop they deliberately don't forward.

**`chore(board)` — record the measured cost of the `/changes` delta poll.** No behaviour change. A backlog card proposed adding ETag/304 to the delta endpoint and a ref-counted poller. Measured first: a hidden tab now makes 0 requests/min (an earlier fix already handles that), a foreground board 12, 24 with a card modal, 36 with settings — dropping to a sixth once the user's first edit confirms push. A 304 would save 6ms of a 45ms request; the other 39ms is framework bootstrap plus the ACL gate, which a conditional request still pays. Neither half was built; the numbers and the reasoning are recorded next to the claim they correct, so it isn't re-proposed without a workload where that 6ms is what hurts.

## Correctness and cost, measured rather than assumed

Three cards in this PR were resolved **by measuring and then not building** — each one recorded next to the claim it corrects, so it isn't re-proposed:

- **The `/changes` delta poll.** A backlog card proposed an ETag/304 path and a ref-counted poller. Measured: a hidden tab makes 0 requests/min (an earlier fix already covers that), a foreground board 12, 24 with a card modal, 36 with settings — dropping to a sixth once the user's first edit confirms push. A 304 would save 6ms of a 45ms request; the other 39ms is framework bootstrap plus the ACL gate, which a conditional request still pays. Neither half built.
- **The anonymous public-share read.** Proposed: throttle the success arm and add an ETag. Both declined with evidence. `#[BruteForceProtection]` is a *failure counter*, not a rate limiter — throttling success makes the 5th view from one address sleep ~3s and the 11th in half an hour a flat 429, firing on any NAT'd office. And an ETag keyed on the change log would be **wrong**: renaming a mentioned member changes the anonymous payload while the board's latest change id stays put. Mention resolution measured at **+5ms** on a realistic board.
- **The purge table lists.** The per-card and board cascades were compared entry by entry before being collapsed; they matched 1:1, including a relations table swept by an `OR` over two columns and recur rules keyed by `template_card_id`. Nothing was hiding in the duplication — but the 15 now-unused mapper methods are deleted, so the old path can't be quietly re-adopted.

## Additional hardening

**`fix(trash)` — clear a card's notifications when it is permanently deleted.** The board purge cleared them; the far more common per-card purge did not, leaving rows that no handler ever deletes and that keep inflating the unread badge. The sweep runs *first*, before anything destructive, because the checklist ids that key `step_assigned` notifications are gone after the rows are deleted — and a failure leaves the card whole in the trash rather than half-deleted.

**`fix` — say when an action was rate-limited.** Nextcloud's rate-limit middleware returns an empty body, so its "Rate limit exceeded" text never reaches the wire and every limited action showed generic failure text — inviting the retry that fails again. The error string is duplicated across ~180 call sites, so the fix went one layer down: a response interceptor at the single axios choke point, covering all seven rate-limited endpoints with zero call-site edits.

**`fix` — show Kanso's own icon where ImageMagick can't read SVG.** `img/` shipped only an SVG, so those instances fell back to the core Nextcloud logo. Verified by removing the SVG coder and confirming all three states, including the counterfactual.

**`refactor(realtime)` — one move-pending guard on the delta path, pinned.** The same defensive check existed in four places; duplicated guards are individually unfalsifiable, and deleting the load-bearing one left all 8 realtime tests green. The new test has to start a drag *from inside the in-flight request* to reach it. The redundant copy is gone.

**`test(e2e)` — the push path and the partial-failure path.** Realtime push had **zero** CI coverage (the positive test only ever ran on a developer's machine). A new non-required `e2e-push` job covers it, plus a global-setup guard that aborts with one clear message when push is advertised but dead — instead of six mystery reds 40 minutes in. And the bulk partial-failure toast, the property that justified shipping "Archive all cards" without a confirm dialog, is now pinned in both its consumers.

## Review notes

- **The `e2e-push` job has never run on a real runner** — that is exactly why it is not a required check. Add it to required checks only after a few green runs, mirroring the existing `upgrade` precedent.
- A purged board's **Activity** entries still show card titles until `activity_expire_days` (default 365) ages them out. Nextcloud exposes no supported per-object delete, so this is documented in the purge service with the evidence rather than worked around.
- The three `NcCheckboxRadioSwitch` instances under board settings → General render as checkboxes while the same component is a toggle elsewhere in that modal.
- Multi-select remains flat-board only — swimlane and List views never passed selection props down.

Closes #146
Closes #147


aktasfatih:
:tada: This pull request is included in version 0.33.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.33.0).

PR #153, opened 2026-09-17
Public-share description images, GitHub label sync, and two interaction fixes

### LINKED ISSUE #152: Images attached to the description of a card in public shares are not displayed
### What happened?

Images attached to the description of a card in public shares are not displayed

### Steps to reproduce

Internal board:
Image

Public shared board:
Image

### Kanso version

0.33.0

### Nextcloud version

34.0.4

### Database

None

### Errors

```shell

```

### Before submitting

aktasfatih:
Four cards from the 2026-09-17 board session: one user-reported bug on the public-share surface, the answered GitHub label-sync feature, and two interaction fixes.

## The public-share image bug is older than it looked

**`fix(sharing)` — show a card's description images on a public shared board** (#152)

Reported against 0.33.0, on a path the previous PR touched — so the first question was whether we regressed it. **We didn't.** It has been broken since **v0.11.0** (Aug 2026), the release where public links first rendered descriptions as markdown.

Three observations settled it rather than reasoning: the `` element **is** present, visible, `max-width: 100%` — so CSS was never involved; it renders 34.95 × 21px with `naturalWidth = 0`, the browser's broken-image box; and the request behind it returns **401 on `/apps/kanso/api/cards/1/attachments/1/inline`**, an authenticated route being hit by a visitor with no session.

The anonymous payload now re-points those srcs — payload-only, like the existing mention redaction — at a new token-gated route, behind: token validity (honouring expiry/disable/rotate) → board-scoped card lookup in the anonymous public-only scope → stack-not-archived → **the card's own public text must actually embed that attachment** → the same IDOR guard and raster-only allow-list as the authenticated route.

**A code review caught two real security defects in the first cut**, both fixed and covered:
- **Token exfiltration.** The src-rewrite regex was unanchored, so it matched the path *inside an absolute URL*. Any EDIT member — who cannot read the share token, since `getConfig` is MANAGE-only — could write `[click](https://evil.example/apps/kanso/api/cards/1/attachments/2/inline)` and have the payload splice the 64-char token into a live external link. Five hostile shapes reproduced, fixed with a left boundary, pinned by a dataprovider.
- **Throttle DoS.** Counting brute-force attempts for *attachment* misses meant 20 broken image references in one description would 429 every visitor behind that IP off the whole board.

Cross-board denial is proven twice — PHPUnit, plus a real-DB e2e fetching board B's attachment with board A's token (404, no bytes). Four mutations confirm the guards aren't vacuous.

## GitHub label sync, to the owner's spec

**`feat(webhooks)` — mirror GitHub issue label changes onto the linked cards**

A `labeled` delivery adds the board label matching the payload's **top-level `label.name`** (case-insensitive) to every linked card; `unlabeled` removes it. Top-level rather than `issue.labels` is load-bearing — on a removal the label is absent from that array, so it cannot express the delta.

Routed through `LabelService::assign/unassign` as the board owner, so the EDIT gate, visibility guard, cross-board rejection and the `kanso_changes` row all come for free. The change-log claim is proven **live, not by mock**: the e2e reads `GET /cards/{id}/activity` and asserts verb 6 with `detail.to = 'Bug'` and verb 7 with `detail.from = 'Bug'`.

Three constraints from the owner, each pinned by a mutation:
- **Never auto-creates a board label.** An unmatched name is a silent no-op. Auto-creating would let anyone with push access to the linked repo mint labels through an unauthenticated endpoint, since creation is MANAGE-gated and the webhook impersonates the owner.
- **Issues only — pull requests change nothing.** That test is non-vacuous for a subtle reason: `changedLabel` *is* populated on the PR path, so it proves the path deliberately ignores it rather than passing on absent data.
- Intake-copies-existing-labels stays out of scope.

**Forgejo is deliberately NOT wired.** The first cut wired it and added three Forgejo tests built from GitHub-shaped payloads — dead code under green, vacuous tests. Forgejo spells it `label_updated`/`label_cleared` and ships no top-level `label`, only the full post-change set, so mirroring it means diffing against card state: different semantics, tracked separately. The replacement test asserts a real `label_updated` delivery mirrors nothing, mutation-proved so a future half-wire fails loudly.

## Interaction fixes

**`fix(search)` — Escape now leaves the search box so board shortcuts work again.** The caret stayed in the input, so every single-key shortcut was swallowed as typing. Escape already *cleared* the query and a spec pinned that, and both neighbours (command palette, card modal) treat Escape as cancel/dismiss — so it stays clear-and-blur rather than changing shipped behaviour. **No new listener was added**: the handler stays on the input, no document listener, no capture hook. The Escape-ownership regression suite from the previous release was re-run green.

**`fix(card)` — line the card reference up with its title.** Not an undefined class (checked first, per the recent precedent) but an adjacent failure: `.card-modal__ref` declares `height: 20px`, but it's a `<button>`, so Nextcloud core's `--default-clickable-area` min-height wins and it renders **34px**. The row then top-aligned and compensated with a `margin-top: 6px` tuned against the *intended* 20px and one line count. Measured before: centres 7.94px apart. After: 0.0px on one line and on two.

The mutation is the interesting part — reintroducing `flex-start` plus a margin hand-tuned for one line makes the one-line case **pass** and the two-line case fail by 14.06px, which is exactly the failure mode the card warned about.

Six surfaces render an id/title pairing; all six were checked, one had the defect. Public-share has the same *shape* of fudge but not the defect, and its row also holds the close ×, which centring would drag to the middle of a long title — left alone deliberately.

## Notes for review

- **Nextcloud 35 support was attempted and correctly refused.** Every `nextcloud:35*` tag on Docker Hub is an empty manifest index — `"manifests":[]`, `full_size: 0` — because Official Images stamps the tag record before the build publishes layers. `docker pull` fails with `no matching manifest for linux/amd64`. No config was changed: `max-version=35` without ever booting 35 would put an unverified support claim on the App Store. Kanso stays hidden on NC 35 until upstream publishes real layers.
- The card reference chip renders 34px rather than the 20px its CSS asks for, because NC core's button min-height wins. Left as-is — shrinking it would fight the touch-target minimum, and the card asked for alignment, not a resize.
- `no_label_match` is a one-bit oracle over a board's label vocabulary, probeable by whoever holds the webhook secret. Inherent to the diagnosable-reason design; recorded honestly in the docblock rather than claimed equivalent to the intake precedent.

Closes #152


aktasfatih:
:tada: This pull request is included in version 0.34.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.34.0).

PR #154, opened 2026-09-18
Show a readable text excerpt on public board tiles

aktasfatih:
Public board tiles rendered the card description as **raw markdown source**, interpolated as plain text:

```html
<p v-if="card.description" class="public-card__desc">{{ truncate(card.description) }}</p>
```

So a tile literally showed ``.

This was never image-specific — every markdown construct leaked as source (`**bold**`, `# headings`, `[label](url)`, code fences). Images were simply the most visible case, because an inline-attachment URL is long enough to consume the whole truncated excerpt, so the tile showed a URL instead of any content.

## What changed

`markdownToPlainText()` in `src/services/markdown.js` flattens a description to plain text, and `PublicBoard.vue` truncates **after** flattening rather than before — otherwise a stripped-away URL still spends the character budget. A description that flattens to empty (image-only) renders no excerpt element at all, rather than a blank paragraph.

**Images are dropped entirely**, not replaced by alt text: an alt of `image.png` is noise, and a tile excerpt is a text summary where an image contributes nothing.

## Derived from the parser, not regexes

The flattener walks `md.parse()` token output from the **same markdown-it instance `renderMarkdown()` uses**, rather than hand-rolling a second set of patterns that would drift from the renderer. That means it already understands nested emphasis, reference links, fences, and the app's own `kanso_mention` / `kanso_cardref` rules for free. `image` tokens are skipped without descending into their alt children; block boundaries become a space; whitespace collapses.

## Rendering images on tiles was considered and rejected

Not a rendering bug to fix later — a deliberate call:

- Tile height becomes unpredictable on a virtualized, column-scanning surface.
- An arbitrary mid-description image would become the card's visual identity, which is what makes Trello-style *chosen* covers different.
- On a public board every tile image is another anonymous request through the token-gated attachment route (token validation, board scoping, embedded-reference check) — a twelve-card column would fire twelve on load, where the authenticated board fires none.
- The **internal board shows no description on tiles at all** (the summary-payload performance bet), so rendering images only on the public view would make the public board richer than the app itself.

If card images are ever wanted, the coherent form is an explicit cover image on both surfaces — cards already carry `coverColor` — not markdown rendering inside a tile.

## Verification

Four new e2e cases in `tests/e2e/public-share.spec.js`, under a block that opens with `test.use({ storageState: { cookies: [], origins: [] } })` so they run as a genuine anonymous visitor rather than false-passing as the shared admin: image dropped with surrounding prose intact, image-only → no excerpt element, bold/link/heading/list/code read as text, and truncation applied to the stripped text. Whole file re-run: **30/30**.

**Two mutations, each killing a different assertion:** restoring `{{ truncate(card.description) }}` fails all four; swapping the order to `markdownToPlainText(truncate(...))` fails **exactly** the truncation-ordering test and leaves the other three green — so that ordering assertion is provably not redundant.

JS unit 161 pass. `l10n:check` clean. A real public share loaded as an anonymous context showed the excerpt reading `"Prose before. Prose after."` with zero console errors or warnings. No PHP touched.

## Noted, deliberately not done

The public board payload still ships **every card's full description** to anonymous visitors in order to render a ~240-character snippet. Client-side flattening doesn't change that. Worth a follow-up — truncate server-side, or stop sending descriptions in the public board payload entirely, as the internal board already does.

`markdownToPlainText` has no unit test: `src/services/markdown.js` calls `DOMPurify.addHook()` at module load, which throws under plain node, and the repo has no jsdom devDependency. Adding one (or restructuring security-sensitive hook registration for testability) was judged scope creep; behaviour is pinned by the four Playwright assertions instead.


aktasfatih:
:tada: This pull request is included in version 0.34.0 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.34.0).

PR #160, opened 2026-09-20
fix: refuse to install on a PHP version too old to run Kanso

### LINKED ISSUE #158: Unable to load/create boards
### What happened?

Kanso is installed but cannot be used due to a bug.

### Steps to reproduce

I am using Kanso 0.34.0 with Nextcloud 32.0.15. When displaying boards or when creating a new board i get the error message "Boards konnten nicht geladen werden. Bitte erneut versuchen." (Translated "Unable to load boards, Please retry.").

The request to https://mydomain/nextcloud/index.php/apps/kanso/api/boards returns a http error 500 and the error text describes that it is necessary to enable javascript.
(although its enabled in my browser).

I am using the latest Firefox and Chrome on Linux.

### Kanso version

0.34.0

### Nextcloud version

32.0.15

### Database

MySQL / MariaDB

### Errors

```shell
Error 500
```

### Before submitting

aktasfatih:
Fixes #158.

## The bug

Kanso's shipped tree uses PHP 8.2 syntax — `readonly class` in `lib/Access/ViewerContext.php` — but `appinfo/info.xml` declared no PHP floor, only the Nextcloud range. Nextcloud 32 still runs on PHP 8.1, so an 8.1 server had nothing stopping it: the App Store offered the app, `occ app:enable` accepted it, and every request that loaded that class died before any Kanso code ran:

```
Exception: syntax error, unexpected token "readonly", expecting "abstract" or "final" or "class"
in file '.../apps/kanso/lib/Access/ViewerContext.php' line 26
at lib/private/AppFramework/Http/Dispatcher.php line 150
```

The reporter (PHP 8.1.2 + NC 32.0.15) saw only a bare HTTP 500 on the boards list, with nothing pointing at PHP.

## The fix

- `appinfo/info.xml`: `<php min-version="8.2"/>` (first in `<dependencies>`, per the App Store schema's `xs:sequence`)
- `composer.json`: matching `"require": { "php": ">=8.2" }` — its `config.platform` already pinned 8.2.27, so only the declaration was missing
- `composer.lock`: content-hash + platform refreshed via `composer update --lock` (no package changes)

No `max-version`: a ceiling would strand servers on a newer PHP that works fine.

## Verification

Mutation-tested end to end on a throwaway NC 32 stack (the app tree mounted live):

```
# with an impossible floor injected
$ occ app:enable kanso
App "Kanso" cannot be installed because the following dependencies are not fulfilled: PHP 9.9 or higher is required.

# with the real floor
$ occ app:enable kanso
kanso 0.34.0 enabled
$ curl .../api/boards  ->  HTTP 200
```

`node scripts/validate-info-xml.mjs` passes (App Store schema), `composer validate` passes.

## Why CI never caught it

`unit-php` runs 8.2/8.3, psalm's platform is pinned to 8.2.27, and every `nextcloud:*` image ships PHP 8.3 — the one version that breaks is the one nothing executes. The declaration itself is the guard: Nextcloud now refuses the install and says why.

## Summary by CodeRabbit

* **Compatibility**
  * The minimum supported PHP version is now 8.2.
  * Installation and enablement are blocked on PHP versions below 8.2.


aktasfatih:
:tada: This pull request is included in version 0.34.1 :tada:

The release notes are on [GitHub](https://github.com/aktasfatih/kanso/releases/tag/v0.34.1).

